PR #3, opened 2017-06-02
Extract common webpack configuration (DPT-495)

JeffreyMFarley:
This is a preliminary commit, will require more testing to verify this approach is correct

@Scotchester @sephcoster


sephcoster:
👍  so far. Merge approach looks good.

sephcoster:
👍

PR #5, opened 2017-06-07
Provide a thin Django implementation for the UI

JeffreyMFarley:

## Additions

- The `ccdb5_ui` directory contains a thin Django implementation that allows it to be used as a plugin for `cfgov-refresh`

## Removals

- `dist/index.html`.  This will not be needed whether deployed as a library or within `cfgov-refresh`
- The open source checklist

## Changes

- Updated the webpack to ignore building font assets
- Installation instructions

## Testing

- None added

## Review

- @sephcoster @Scotchester @willbarton 

## Notes

- ~~Why Maria? [She helped Django!](http://www.imdb.com/title/tt0060315/plotsummary?ref_=tt_stry_pl)~~

## Todos

- Integrate with `cfgov-refresh`

## Checklist


sephcoster:
This looks like a solid start to me 👍  

Probably want @willbarton @rosskarchner or @Scotchester to weigh in regarding any specific settings that play nicely with the eventual install.

rosskarchner:
I think the big missing thing is a setup.py, that would allow us to generate a wheel that includes the python code and the built front-end assets.

especially ensure that setup_requires includes cfgov_setup, and frontend_build_script points to a shell script that builds the frontend. For example: https://github.com/cfpb/college-costs/blob/master/setup.py#L59-L60

When you're done, running `pip wheel .` in the root directory should generate a .whl file that includes the django code and front end assets.

willbarton:
You're going to want to separate the ccdb5-specific views and templates into a Django app instead of a project. With the set up you have now, that'll require including that app in `maria`'s `INSTALLED_APPS`. 

But with @rosskarchner's suggested `setup.py`, you could also simply `pip install -e` into the same virtualenv as cfgov-refresh and run as part of cfgov-refresh and not really need the independent Django project here.

JeffreyMFarley:
@rosskarchner Do I really need to build the front-end assets?  They are already built in the `dist` folder and would get included with a `git clone`

Scotchester:
> Do I really need to build the front-end assets?  They are already built in the `dist` folder and would get included with a `git clone`

@JeffreyMFarley I'm still wondering why we're checking in the `dist` folder :)

JeffreyMFarley:
@Scotchester 
1. Because I don't think any consumers of the library should have to build
1. Because you can use Github as a CDN by using `rawgit`
1. Because the library developer should be the one with the most knowledge of how to build the library

What is your counterargument?

JeffreyMFarley:
@rosskarchner @willbarton @Scotchester @sephcoster I have adjusted the code to meet the concerns you have raised. 

Following the example from https://github.com/cfpb/college-costs,
* I split the project part from the application part
* I liked how the `STANDALONE` variable was used, so I incorporated (stole) that
* I used the `setup.py` as a template to build one applicable to this repo

I have also verified that this version of the install brings over the necessary files when executing `pip install`

sephcoster:
Thanks again for your feedback all - looks like we're good to go. 👍

PR #15, opened 2017-06-27
Checkbox dispatchers

sephcoster:
Filter checkboxes dispatch appropriate actions to call reducers to adjust query state to represent the newly selected values in the query.

## Additions

- FILTER_CHANGED action
- Query reducer updates as a result of filter change action
- AggregationItem component using redux / connect

## Removals

-

## Changes

- Query reducer code was changed with the addition of the filter change function

## Testing

- Coverage remains at 100%.

## Review

- @JeffreyMFarley 

[Preview this PR without the whitespace changes](?w=0)

## Notes

-

## Todos

- Update default state to first check URL structure so state remains the same on page refresh and links can be sent to others to duplicate the filter criteria.

## Checklist

* [ X ] Changes are limited to a single goal (no scope creep)
* [ X ] Code can be automatically merged (no conflicts)
* [ X ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ X ] Passes all existing automated tests
* [ X ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ X ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #17, opened 2017-06-28
[WIP] General filter work

JeffreyMFarley:
This pull request accomplishes a few things:
* Integrates with the API as of PR https://github.com/cfpb/ccdb5-api/pull/7
* Establishes the order/position for all filters
* Begins the work on "See X more results"
* Implements "Show/Hide" on the filter 
* `Clear All` action
* `Remove Filter` action
* Default state from URL [example](http://localhost:3000/?from=0&product=Debt%20collection&searchText=&size=10&sort=relevance_desc)

## Testing

- Testing remains at 100%

## Review

- @sephcoster  @kurzn 

## Screenshots

screen shot 2017-06-28 at 2 59 47 pm

## Todos

- [ ] `Show more` logic

## Checklist

### DIFF
--- a/src/reducers/query.jsx
+++ b/src/reducers/query.jsx
@@ -10,7 +10,10 @@ export const defaultQuery = {
 const urlParams = ['searchText', 'from', 'size'];
 const urlParamsInt = ['from', 'size'];
 
-export function processParams(state, params) {
+// ----------------------------------------------------------------------------
+// Complex reduction logic
+
+function processParams(state, params) {
   const processed = Object.assign({}, state)
 
   // Filter for known
@@ -65,6 +68,20 @@ export function toggleFilter(state, action) {
   }
 }
 
+function removeFilter(state, action) {
+  const newState = {...state}
+  if (action.filterName in newState) {
+    const idx = newState[action.filterName].indexOf(action.filterValue)
+    if (idx !== -1 ) {
+      newState[action.filterName].splice(idx, 1)
+    }
+  }
+  return newState
+}
+
+// ----------------------------------------------------------------------------
+// Action Handler
+
 export default (state = defaultQuery, action) => {
   switch(action.type) {
   case types.SEARCH_CHANGED:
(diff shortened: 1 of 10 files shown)


src/FilterPanel.jsx
```diff
@@ -31,13 +31,52 @@ export class FilterPanel extends React.Component {
...
+        </CollapsibleFilter>
+        <hr />
+        <CollapsibleFilter title="Zip Code"
+                           desc="The mailing ZIP code provided by the consumer">
+            <input type="text" placeholder="Enter first three digits of zip code" />
+        </CollapsibleFilter>
+        <hr />
         <Aggregation title="Did company provide a timely response?"
-                         options={this.props.aggs.timely_response}
+                     options={this.props.aggs.timely}
```
sephcoster: My current plan with #15 is to use a `fieldName` prop that passes the name of the aggregation field as we need that to update params: `fieldName='timely'`

Any issues adding that to `Aggregatin` moving forward or is there a better way to derive it from the component with the current props structure that I should be using?

kurzn:
The input field for company name (and all other filters with that UI element) should be full width of the column.

Also make sure there is 15px both above and below the horizontal rule dividing aggregates from each other.

src/FilterPanel.jsx
JeffreyMFarley: That's fine

src/reducers/query.jsx
```diff
@@ -53,7 +68,20 @@ export function toggleFilter(state, action) {
   }
 }
 
-// TODO: Set defaultQueryState to recognize existing URL params
+function removeFilter(state, action) {
```
sephcoster: FILTER_CHANGE also does pill removal in a way (detects if already present then filters from the returned array, rather than splicing like this approach). For future readability, do you think we should differentiate when / how to use filter removal / filter change or are comments the best venue for now?

sephcoster:
👍

src/reducers/query.jsx
JeffreyMFarley: We can refactor for consistency.  I like the idea of a toggle action so I want to keep it.  Maybe we just figure out how to name it better

PR #22, opened 2017-07-06
Typeahead

JeffreyMFarley:
Typeahead!

## Testing

- Testing remains at 100%

## Review

- @sephcoster @kurzn 

#### One character typed
screen shot 2017-07-07 at 4 07 40 pm

#### Two characters typed
screen shot 2017-07-07 at 4 08 01 pm

#### Less than 5 results
screen shot 2017-07-07 at 4 08 27 pm

#### Arrow key selecting an option
screen shot 2017-07-07 at 4 08 45 pm

#### Arrow key selecting one item, mouse hover over another
screen shot 2017-07-07 at 4 09 01 pm

#### After the user has selected an option
screen shot 2017-07-07 at 4 09 18 pm

## To Dos (in another PR)

- [ ] Support for asynchronous callbacks (like when calling the suggest endpoint)
- [ ] Enter and Tab will work differently when the typeahead is used with the search suggester.  Currently, the user _must_ select an option. Free form values are not allowed
- [ ] May need an option to not clear the results after selecting

## Checklist

### DIFF
--- a/src/Typeahead/index.jsx
+++ b/src/Typeahead/index.jsx
@@ -255,7 +255,7 @@ export default class Typeahead extends React.Component {
                   onOptionSelected={this._selectOption.bind(this)}
                   renderOption={this.props.renderOption}
                   selectedIndex={this.state.selectedIndex}
-                  footer="Continue typing to narrow the set of results"
+                  footer="Continue typing for more results"
         />
     )
   }
--- a/src/Typeahead/Typeahead.less
+++ b/src/Typeahead/Typeahead.less
@@ -24,15 +24,16 @@
       left: 0;
 
       li.footer {
-        padding-top: @gutter-narrow;
-        padding-left: 0;
+        padding: @gutter-normal;
+        border-top: solid 1px @gray-40;
         margin: 0;
 
         background-color: @gray-10;
-        color: @pacific;
         cursor: default;
+        font-size: @size-vi;
         font-style: italic;
         list-style-type: none;
+        text-align: center;
       }
     }
   }
--- a/src/Typeahead/__tests__/__snapshots__/Typeahead.spec.jsx.snap
+++ b/src/Typeahead/__tests__/__snapshots__/Typeahead.spec.jsx.snap
@@ -133,7 +133,7 @@ exports[`component::Typeahead render phases renders the TOO_MANY phase 1`] = `
       </li>
       <li
         className="footer">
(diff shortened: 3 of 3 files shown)


src/Filters/Issue.jsx
```diff
@@ -32,11 +47,65 @@ export class Issue extends React.Component {
...
+      key: x.key,
+      normalized: normalize(x.key)
+    }
+  })
+
   return {
-    options: state.aggs.issue
+    options,
+    forTypeahead
   }
```
sephcoster: Is the plan for the ToDo on line 92 to add a MapDispatchToProps with a variant on our existing `FILTER_CHANGE` actions? It occurs to me we may also need a `TOGGLE_PARENT_FILTER` action as well to accomplish this.

src/Typeahead/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+import React from 'react'
+import PropTypes from 'prop-types'
+import Selector from './Selector'
+import * as keys from '../constants'
+import './Typeahead.less'
+
+// ----------------------------------------------------------------------------
+// State Machine
+
+// TODO: Use with onInputChangeAsynch
```
sephcoster: This is a cool way to approach all the edge cases / possible options for this.

src/Typeahead/index.jsx
```diff
@@ -0,0 +1,287 @@
...
+    return (
+      <section className="typeahead"
+               onBlur={this._onBlur}
+               onFocus={this._onFocus}>
+        <input type="text"
+               disabled={this.props.disabled}
+               onChange={this._valueUpdated}
+               onKeyDown={this._onKeyDown}
+               placeholder={this.props.placeholder}
+               value={this.state.inputValue}
```
sephcoster: Maybe here's a better place to set the top-level stuff and pass down so you can change it per distinct type-ahead instance (search Dispatch vs. Issue dispatch, etc)

src/Filters/Issue.jsx
JeffreyMFarley: I knew we had to refactor the parent/child selection logic, so I figured I would handle that TODO when I update the logic

src/Typeahead/index.jsx
JeffreyMFarley: Yeah, if this code needs to be extended we can look into that

PR #23, opened 2017-07-10
Finish Issue Filter

JeffreyMFarley:
This pull request contains the remaining functionality that was left to implement in the Issue filter

## Additions

- Issues that are selected are sorted in front of the non-selected issues
- A filter can be selected from a typeahead

## Changes

- The logic when deselecting a parent

## Testing

- Test coverage remains at 100%

## Review

- @sephcoster @kurzn 

#### Selected options go first
screen shot 2017-07-10 at 1 50 38 pm

#### Selecting a filter from typeahead
typ

## Checklist

### DIFF
--- a/src/Filters/Issue.jsx
+++ b/src/Filters/Issue.jsx
@@ -5,7 +5,7 @@ import AggregationBranch from './AggregationBranch'
 import CollapsibleFilter from './CollapsibleFilter'
 import Typeahead from '../Typeahead'
 import { addMultipleFilters } from '../actions/filter'
-import { normalize, sortSelThenCount } from './utils'
+import { normalize, slugify, sortSelThenCount } from './utils'
 
 export class Issue extends React.Component {
   constructor(props) {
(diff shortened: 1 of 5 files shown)


src/Filters/Issue.jsx
```diff
@@ -88,14 +88,44 @@ export class Issue extends React.Component {
...
+  allIssues.forEach(x => {
+    const idx = x.indexOf(SLUG_SEPARATOR)
+    const key = (idx !== -1) ? x.substr(0, idx) : x
+    if (selections.indexOf(key) === -1)  {
+      selections.push(key)
+    }
+  })
+
+  // Make a cloned, sorted version of the aggs
+  const options = sortSelThenCount(state.aggs.issue, selections)
```
sephcoster: This resembles something that might be done in a reducer. Does it makes sense to keep it here vs. put it in the reducer that is handling the agg state?

JeffreyMFarley: Maybe! 

I put it here because this implementation definitely requires the sort, where the reducer may have to serve many different implementations.

It's a bit of an intellectual exercise since there is only one implementation needed in the application. So it works here or the reducer.

PR #40, opened 2017-07-31
Search bar

JeffreyMFarley:
Implement the SearchBar according to [GHE]/data-products-team/CCDB5/issues/2

Search bar entry
- [x] User starts typing, placeholder text disappears
- [x] After 2 characters are entered, auto-suggest menu appears below search bar
- [x] User can select from the auto-suggest options or keep typing
- [x] Search action executed with "enter" key 
- [x] or "search" button

## Additions

- Updated the typeahead for use by the search bar
  - Asynchronous call for options
  - Allow the user to choose any text, not necessarily one of the options

## Changes

- Focus/Blur does not have separate states, but uses a conditional flag

## Testing

- 💯 

## Review

- @sebworks @wpears

## Screenshots

screen shot 2017-08-01 at 12 52 22 pm

## Todos

- [x] Integrate with API suggest endpoint when it becomes available
- [x] Provide a portal for the waiting, error and no results state
- [x] Debounce the input
- [x] Render-only component for highlighted results

## Checklist

### DIFF
--- a/src/Typeahead/__tests__/Typeahead.spec.jsx
+++ b/src/Typeahead/__tests__/Typeahead.spec.jsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import renderer from 'react-test-renderer'
 import { shallow } from 'enzyme';
-import Typeahead, { MODE_OPEN } from '..'
+import Typeahead, { MODE_OPEN, debounce } from '..'
 import * as keys from '../../constants'
 
 function setupEnzyme(initalProps={}, removeDebounce=true) {
@@ -330,5 +330,33 @@ describe('component::Typeahead', () => {
         expect(target.state('phase')).toEqual('CHOSEN')
       })
     })
+  })
+})
+
+describe('debounce', () => {
+  beforeEach(() => {
+    jest.useFakeTimers()
+  })
+
+  it('calls the passed in function after N milliseconds', () => {
+    const spy = jest.fn()
+    const target = debounce(spy, 200)
+    target()
+    expect(spy).not.toHaveBeenCalled()
+    jest.runTimersToTime(200)
+    expect(spy).toHaveBeenCalled()
+  })
+
+  it('only triggers one call while the timer is active', () => {
+    const spy = jest.fn()
+    const target = debounce(spy, 200)
+
+    target()
+    target()
+    target()
+
+    expect(spy).not.toHaveBeenCalled()
+    jest.runAllTimers()
+    expect(spy).toHaveBeenCalledTimes(1)
   })
 })


wpears:
This ready for :eyes:?

JeffreyMFarley:
@wpears It is now!

src/SearchBar.jsx
```diff
@@ -1,66 +1,168 @@
...
-    };
+      inputValue: props.searchText,
+      searchField: props.searchField
+    }
 
     // This binding is necessary to make `this` work in the callback
     // https://facebook.github.io/react/docs/handling-events.html
-    this._handleSubmit = this._handleSubmit.bind(this);
-    this._updateInputValue = this._updateInputValue.bind(this);
+    this._handleSubmit = this._handleSubmit.bind(this)
```
sebworks: It might be nice to have a `bindAll` utility function. https://www.npmjs.com/package/lodash.bindall

src/SearchBar.jsx
```diff
@@ -1,66 +1,168 @@
...
-            <input type="button" className="btn" value="Types" />
-            <input type="text"
-                   value={this.state.inputValue}
-                   onChange={this._updateInputValue}
-                   placeholder="Enter your search term(s)"
-                   id="searchText" />
-            <input type="submit" className="btn primary" value="Search" />
+            <h5>Search Within</h5>
+            <div className="layout-row">
+              <div className="cf-select">
```
sebworks: The `ask-cfpb` [page](https://www.consumerfinance.gov/ask-cfpb/category-student-loans/) is using a type-a-head as well. 

screen shot 2017-08-02 at 4 50 33 am


It probably makes sense to piggy back on that markup that markup

src/SearchBar.less
```diff
@@ -1,21 +1,57 @@
 @import (less) "base.less";
 
 .search-bar {
```
sebworks: Consider using the ask-cfpb type-a-head styles.

src/Typeahead/HighlightingOption.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+export const HighlightingOption = ({ label, position, value }) => {
```
sebworks: I would use defaults here, if you are assuming they are strings below.

sebworks:
I would add some logic to display notifications on API failures.

src/SearchBar.jsx
JeffreyMFarley: Also: https://stackoverflow.com/questions/32192682/react-js-es6-avoid-binding-this-to-every-method

src/SearchBar.jsx
JeffreyMFarley: @kurzn provided the styling in [GHE]/data-products-team/CCDB5/issues/2

JeffreyMFarley:
@sebworks there already is logic to handle API errors:

If the error happens: https://github.com/cfpb/ccdb5-ui/pull/40/files#diff-79abbaf5643975fca1874849f9092040R227

The error is rendered: https://github.com/cfpb/ccdb5-ui/blob/master/src/Typeahead/index.jsx#L225

The test: https://github.com/cfpb/ccdb5-ui/blob/master/src/Typeahead/__tests__/Typeahead.spec.jsx#L37

src/Typeahead/HighlightingOption.jsx
JeffreyMFarley: I'm not sure what you mean here.  I don't think there are any appropriate default values and the PropTypes section below indicate that the three params are required.

What was your thinking on this?

src/SearchBar.jsx
sebworks: @JeffreyMFarley, I would talk to her about this.

src/SearchBar.jsx
sebworks: So the lodash method allows you to pass in an array of method names. Decorators also might not be a bad idea either.

src/Typeahead/HighlightingOption.jsx
sebworks: ok, I didn't realize you could guarantee that they would be strings.

sebworks:
>> @sebworks there already is logic to handle API errors:


I didn't have the API setup correctly initially. I received the error below ( which is corrected ):

screen shot 2017-08-02 at 9 58 45 am
 
I didn't receive any error notification via the UI.

JeffreyMFarley:
@wpears Any input?

src/Typeahead/index.jsx
```diff
@@ -7,11 +7,36 @@ import * as keys from '../constants'
...
+    if (!timer) {
+      timer = setTimeout(later, wait)
+    }
+  }
+}
+
+// ----------------------------------------------------------------------------
+// Usage Mode
+
+// The user can enter any text of choose one of the drop down options
```
wpears: **or** choose

PR #32, opened 2017-07-25
Adding basic code for linter process using Fewd standard

sebworks:
Adding basic code for linter process using Fewd standard

## Additions

- Added `scripts/lint.js` to handling the linting process. 
- Added a package-lock.json file. Generated using Node version 8 and npm version 5.3.0.

## Changes

- Modified `.eslintrc` to add the FEWD linting file. I made no changes to this file for react or JSX.
- Modified `package.json` file to use linting file instead of running process inline.

## Testing

- Run `npm run lint`
- Run `npm run lint --fix` and then `git status`. You should see that it tried to fix some of the linter errors. 
- Run `npm run lint ./scripts/lint.js` and verify it just ran on that file.

[Preview this PR without the whitespace changes](?w=0)

## Notes

- It doesn't kick out an error if you pass in a file it cant find, it should. 

## Todos

- Fix some linter errors.
- Tailor the `eslint` file for React.

## Checklist

* [ ] Changes are limited to a single goal (no scope creep)
* [ ] Code can be automatically merged (no conflicts)
* [ ] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [ ] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

Issue #26, opened 2017-07-11
Configure ESLint to use the front-end standard


sebworks:
Modify the linting process to use the front-end linting [file](https://github.com/cfpb/development/blob/master/.eslintrc).  Adjustments might need to be made for React and JSX.

JeffreyMFarley:
> Adjustments might need to be made for React and JSX.

Yep!

sebworks:
@JeffreyMFarley, I can create a PR if you would like.

JeffreyMFarley:
@sebworks I started work on this in a branch https://github.com/JeffreyMFarley/ccdb5-ui/commits/lint.  I wasn't going to work on it until we have more of the functionality finished and I have a chance to adjust our current code base.

I do want to see what you have so far!

sebworks:
@JeffreyMFarley, I have an sand sandwach right now but I'll work on it today.

sebworks:
@JeffreyMFarley, ok I've added a basic process: #32. I didn't merge in the modifications you made to the linter file on your branch though.

JeffreyMFarley:
There are some warnings to clean up, but the core of this issue has been addressed

Issue #25, opened 2017-07-11
Upgrade NPM dependencies


sebworks:
I would consider upgrading the NPM dependencies, especially Webpack, to the latest versions .  There are some great features in Webpack 2.x and 3.x. https://webpack.github.io/docs/roadmap.html.

JeffreyMFarley:
I tried that, but the LESS plugin is not available for the latest version yet

sebworks:
ok, I'll check it out today.  I think people are starting to upgrade: https://github.com/facebookincubator/create-react-app/issues/2573.

sebworks:
It looks like `less-loader` just added support for webpack 3: https://github.com/webpack-contrib/less-loader/releases/tag/v4.0.5.

JeffreyMFarley:
`2 days ago`.   Well, there it is

sebworks:
@JeffreyMFarley, ok I've added a basic process: #32. I didn't merge in the modifications you made to the linter file on your branch though.

JeffreyMFarley:
Notes on my first attempt to address this issue https://gist.github.com/JeffreyMFarley/2fc381f9b45c804242aea854165ec628

sebworks:
Ok, I'm going to give it go.

PR #44, opened 2017-08-03
Add Linting support

JeffreyMFarley:
Supersedes https://github.com/cfpb/ccdb5-ui/pull/32

Work towards #25 #26

## Additions

- Lint Rules

## Changes

- Lots

## Testing

- 💯 

## Todos

- Take care of the 175 warnings as modules are being worked

## Checklist

### DIFF
--- a/src/Filters/SingleCheckbox.jsx
+++ b/src/Filters/SingleCheckbox.jsx
@@ -3,11 +3,11 @@ import React from 'react';
 export default class SingleCheckbox extends React.Component {
   render() {
     return (
-      <section className='single-checkbox'>
+      <section className="single-checkbox">
         <h5>{this.props.title}</h5>
-        <div className='m-form-field m-form-field__checkbox'>
-            <input className='a-checkbox' type='checkbox' id='theCheckbox' />
-            <label className='a-label' htmlFor='theCheckbox'>{this.props.label}</label>
+        <div className="m-form-field m-form-field__checkbox">
+            <input className="a-checkbox" type="checkbox" id="theCheckbox" />
+            <label className="a-label" htmlFor="theCheckbox">{this.props.label}</label>
         </div>
       </section>
     );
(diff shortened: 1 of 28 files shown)


sebworks:
looks good, 👍

src/Filters/SingleCheckbox.jsx
```diff
@@ -3,12 +3,12 @@ import React from 'react';
 export default class SingleCheckbox extends React.Component {
   render() {
     return (
-      <section className="single-checkbox">
```
sephcoster: How do we feel about single checks for HTML / JSX? Seems like this is a case where you'd stick to double quotes for actual HTML properties but single quotes for JavaScript-related strings for readability but I'm OK either way as long as we're consistent.

src/constants/index.jsx
```diff
@@ -37,17 +37,17 @@ export const knownFilters = [
...
-export const VK_UP                     = 38
+export const VK_DOWN = 40
+export const VK_ENTER = 14
+export const VK_ESCAPE = 27
+export const VK_RETURN = 13
+export const VK_TAB = 9
+export const VK_UP = 38
 
-///////////////////////////////////////////////////////////////////////////////
+/// ////////////////////////////////////////////////////////////////////////////
```
sephcoster: Some weirdness here adding a space on the eslint fix but nothing to necessarily change. I also liked the formatting of constants but not necessarily worth writing a very specific lint rule for that spacing. Might be worth ignoring that file, however to allow for more readable constants if we prefer that format.

src/Filters/SingleCheckbox.jsx
JeffreyMFarley: I agree with you, but not sure how to configure eslint to do that

JeffreyMFarley: I think I found where I can change the HTML/JSX quotes.  Will resubmit because I agree with ypu

PR #45, opened 2017-08-04
Initially hiding non-checked filters

AdamZarger:
Added a SimpleFilter component for checkbox filters. Added logic to expand/collapse filters based on URL parameters on initial page load.

## Additions

- SimpleFilter and SimpleCollapsibleFilter React components
- SimpleFilters are automatically collapsed and will automatically expand if a particular filter is present in the URL parameters

## Testing

- To Do

## Review

- @JeffreyMFarley 
- @amymok 

## Todos

- Fix broken tests
- Write tests for new components

## Checklist

### DIFF
--- a/src/Filters/__tests__/__snapshots__/CompanyName.spec.jsx.snap
+++ b/src/Filters/__tests__/__snapshots__/CompanyName.spec.jsx.snap
@@ -22,11 +22,11 @@ exports[`component::CompanyName snapshots renders without crashing 1`] = `
     The company name as it appears in our complaint system, which may be different than the name the consumer provided in their complaint
   </p>
   <section
-    className="typeahead"
+    className="typeahead "
     onBlur={[Function]}
     onFocus={[Function]}>
     <input
-      disabled={false}
+      autoComplete="off"
       onChange={[Function]}
       onKeyDown={[Function]}
       placeholder="Enter company name"
--- a/src/__tests__/__snapshots__/App.spec.jsx.snap
+++ b/src/__tests__/__snapshots__/App.spec.jsx.snap
@@ -237,11 +237,11 @@ exports[`initial state renders without crashing 1`] = `
             The company name as it appears in our complaint system, which may be different than the name the consumer provided in their complaint
           </p>
           <section
-            className="typeahead"
+            className="typeahead "
             onBlur={[Function]}
             onFocus={[Function]}>
             <input
-              disabled={false}
+              autoComplete="off"
               onChange={[Function]}
               onKeyDown={[Function]}
               placeholder="Enter company name"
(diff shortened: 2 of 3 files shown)


src/Filters/SimpleCollapsibleFilter.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+
+export default class SimpleCollapsibleFilter extends React.Component {
```
JeffreyMFarley: I don't think you need the additional class.  You should be able to fix `CollapsibleFilter`

src/Filters/SimpleCollapsibleFilter.jsx
```diff
@@ -0,0 +1,52 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      showChildren: typeof props.showChildren === 'undefined' ? false : props.showChildren
```
JeffreyMFarley: This line should be `showChildren: props.showChildren`

At the bottom, there should be something like:

```Javascript
CollapsibleFilter.propTypes = {
  showChildren: PropTypes.bool
}

CollapsibleFilter.defaultProps = {
  showChildren: true
}
```

This way, the default behavior is 'shown' but can be overridden, like is done in `SimpleFilter`

src/Filters/SimpleFilter.jsx
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+export const mapStateToProps = (state, ownProps) => {
+  // Find all query filters that refer to the field name
+  const activeChildren = state.query[ownProps.fieldName] || []
+
+  return {
+    options: state.aggs[ownProps.fieldName] || [],
+    showChildren: activeChildren.length > 0
```
JeffreyMFarley: 👍

JeffreyMFarley:
Awesome! Now just get back to 100% coverage and we are good to go

AdamZarger:
My mistake, I didn't realize I had pushed these changes. Thanks for your suggestions! As you saw I was able to implement them without needing the custom filter. I'm working on adapting the UrlBar tests now

src/Filters/__tests__/SimpleFilter.spec.jsx
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+  describe('componentWillReceiveProps', () => {
+    it('changes props correctly', () => {
+      props.params.timely = 'no'
+      const expected = 'no'
+
+      target.componentWillReceiveProps(props)
+
+      expect(target.props.params.timely).toEqual(expected)
```
JeffreyMFarley: I realize you borrowed this code from `UrlBarSynch`, but it won't work here, and that's why you get this error:

```
 Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the SimpleFilter component.
```
This is more like how the test should work:

```javascript
      const {target} = setupEnzyme()
      target.setProps({showChildren: true})
      expect(target.state('showChildren')).toEqual(true)

```

What is `setupEnzyme`? There are a lot of examples in the codebase...

AdamZarger: Thank you for this feedback! I was pretty sure that there was a better way than how I was doing this. I was able to use a different example of `setupEnzyme` to figure out how to properly write this test

PR #50, opened 2017-08-07
Product/Subproduct Filter

amymok:
This is to add a Product/Subproduct filter

## Additions

- Added product filter

## Todos

-

## Checklist


sephcoster:
We will likely want to keep the generalized Aggregation component as part of the repository so we can use it for simpler field aggregations that only require a checkbox / expansion and take the basic parameters such as "How Complaint Submitted" or Consent values.

JeffreyMFarley:
I think I am OK with removing the generalized `Aggregation` class even though it was a good workhorse
1.) No classes currently use it
2.) If we do need classes in the future, they will get remixed from the current components
3.) YAGNI

sephcoster:
Agreed - I missed that "SimpleFilter" components were now handling that stuff so no reason to keep it!

src/Filters/Product.jsx
```diff
@@ -0,0 +1,66 @@
...
+    const listComponentProps = {
+      fieldName: 'product'
+    }
+
+    return (
+      <CollapsibleFilter title="Product / sub-product"
+                         desc="The type of product and sub-product the consumer identified in the complaint"
+                         showChildren={this.props.showChildren}
+                         className="aggregation">
+        <a href="http://files.consumerfinance.gov/f/documents/201704_cfpb_Summary_of_Product_and_Sub-product_Changes.pdf" target="_blank">Recent changes to products and sub-products</a>
```
sephcoster: It looks like this aggregation needs a slightly different style and additional spacing to match the comp. That said, that would not hold up merging this but would require adding a smaller clean-up task for a revisit.

PR #69, opened 2017-08-24
Removed Share Links, copy base_update, namespace templates

### LINKED ISSUE #64: Don't put templates at app root level
The templates for this app are located in the root `ccdb5_ui/templates` level and thus referred to with no subdirectories when rendering, e.g. [`template_name = 'standalone-base.html'`](https://github.com/cfpb/ccdb5-ui/blob/master/ccdb5_ui/views.py#L11).

This works as long as all template names are guaranteed to be unique across all apps in a Django project, but could cause conflicts if, say, another app also has a `standalone_base.html` in its root `templates` directory. Prefixing all template names with `ccdb-` might work but has to be done consistently.

The Django convention seems to be to place templates under subdirectories named for the app, e.g. `ccdb5-ui/templates/ccdb5-ui/main.html`. At first this might feel duplicative but it makes the view code clearer, especially if these templates are ever referred to from another project, e.g. `'template_name = 'ccdb5-ui/main.hml`.

From [the Django tutorial](https://docs.djangoproject.com/en/1.8/intro/tutorial03/#write-views-that-actually-do-something):

> Now we might be able to get away with putting our templates directly in polls/templates (rather than creating another polls subdirectory), but it would actually be a bad idea. Django will choose the first template it finds whose name matches, and if you had a template with the same name in a different application, Django would be unable to distinguish between them. We need to be able to point Django at the right one, and the easiest way to ensure this is by namespacing them. That is, by putting those templates inside another directory named for the application itself.

## Current behavior

Templates for this app live under `ccdb5-ui/templates`.

## Expected behavior

Templates for this app live under `ccdb5-ui/templates/ccdb5-ui`.

sephcoster:

## Additions

- 

## Removals

- Share block from base_update template emptied

## Changes

- Closes #64 by namespacing templates in a folder by app name

## Testing

- 💯 

## Review

- @JeffreyMFarley @AdamZarger 

[Preview this PR without the whitespace changes](?w=0)

## Notes

- Closes DPT-610

## Todos

-

## Checklist


ccdb5_ui/templates/ccdb5_ui/ccdb_base.html
```diff
@@ -0,0 +1,227 @@
...
+        {% endblock %}
+        <div id="wrapper">
+        <section id="maincontent">
+            <ul class="bread meta">
+                {% block breadcrumbs %}
+                {% endblock %}
+            </ul> <!-- /class=bread -->
+                {% block content %}
+                {% endblock %}
+                <!-- {% block share_box %}
```
sephcoster: This is the deprecated share block - required cloning the legacy cfgov `base_update.html` template from legacy and removing.

sephcoster:
I'll check with the team for any recommendations to make sure we stay in sync.

PR #100, opened 2017-10-12
React 16 changes for IE9

JeffreyMFarley:
@sephcoster

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -71,7 +71,7 @@ module.exports = merge( {
         loader: 'string-replace',
         query: {
           search: '@@API',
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -10,7 +10,7 @@
 
     <script src="https://cfpb.github.io/capital-framework/dist/static/js/main.min.js"></script>
     <!--[if IE 9]><!-->
-    <script src="https://cdn.polyfill.io/v2/polyfill.js?features=Intl.~locale.en"></script>
+    <script src="https://cdn.polyfill.io/v2/polyfill.js?features=default,Intl.~locale.en"></script>
```
sephcoster: Can we call `polyfill.min.js` to save some bytes?

JeffreyMFarley: Sure

PR #104, opened 2017-10-17
Clicking the clear button will reset the text

JeffreyMFarley:
...and cause another search to be executed.  If this is not desired behavior additional functions/callbacks will need to be created since the parent (SearchBar) does not have a callback for receiving a changed value that does nothing

* _onInputChange calls the suggest endpoint
* _onOptionSelected calls a search


sephcoster:
That behavior should work!

PR #105, opened 2017-10-17
[WIP] Polyfill Detection

sephcoster:
Short description explaining the high-level reason for the pull request

## Additions

- Updated templates to check browser object support prior to polyfill grab

## Removals

- None

## Changes

- Prior to react initialization, browser detects supported methods for loading polyfills as required

## Testing

- 💯 

## Review

- @JeffreyMFarley 

[Preview this PR without the whitespace changes](?w=0)

## Screenshots

N/A

## Notes

- None

## Todos

- Update method calls to reflect all required methods that might need to be polyfilled.

## Checklist

* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [x] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [x] Passes all existing automated tests
* [ ] New functions include new tests
* [ ] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #109, opened 2017-10-20
Lazy polyfill

JeffreyMFarley:
This changes the implementation from #105 slightly.

Since the code is shifted from the HTML into a JS module, the react/webpack bootstrap routine is run before the `main()` routine is called.  However, the react/webpack bootstrap requires `Set` and maybe `Map`, so those have been included in a polyfill that will always travel in the minified `ccdb5.min.js`.  The feature detection is still there for the `Intl` polyfill

### DIFF
--- a/ccdb5_ui/templates/ccdb5_ui/ccdb-main.html
+++ b/ccdb5_ui/templates/ccdb5_ui/ccdb-main.html
@@ -48,9 +48,6 @@
 {% endblock %}
 
 {% block page_js %}
-  <!--[if IE 9]><!-->
-  <script src="https://cdn.polyfill.io/v2/polyfill.js?features=default,Intl.~locale.en"></script>
-  <!--<![endif]-->
   <script src="{% static "ccdb5.min.js" %}" type="text/javascript"></script>
 {% endblock %}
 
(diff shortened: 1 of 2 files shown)


JeffreyMFarley:
FYI:  Running the development harness in IE9 produces the following in the console:

```
Warning: Failed prop type: Invalid prop `children` supplied to `<<anonymous>>`, expected a ReactNode.
    in Unknown (at App.jsx:87)
    in Unknown (at index.js:41) 
Warning: Failed prop type: Invalid prop `children` of type `object` supplied to `Provider`, expected a single ReactElement.
    in Provider (at App.jsx:85)
    in Unknown (at index.js:41) 
Warning: Failed prop type: Invalid prop `children` of type `object` supplied to `IntlProvider`, expected a single ReactElement.
    in IntlProvider (at App.jsx:46)
    in Unknown
    in Unknown (at App.jsx:89)
    in Unknown (at App.jsx:87)
    in Unknown
    in Unknown (at App.jsx:86)
    in Provider (at App.jsx:85)
    in Unknown (at index.js:41) 
Warning: The tag <summary> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter. 
Warning: The tag <main> is unrecognized in this browser. If you meant to render a React component, start its name with an uppercase letter. 
```

However, everything runs and works

The messages do not appear in the cfgov hosted version

PR #126, opened 2017-11-16
Last updated date

AdamZarger:
Last updated date now keyed off of last_indexed date rather than last_updated date
## Review

- @sephcoster 
- @JeffreyMFarley

### DIFF
--- a/src/SearchPanel.jsx
+++ b/src/SearchPanel.jsx
@@ -7,10 +7,10 @@ import SearchBar from './SearchBar'
 
 export class SearchPanel extends React.Component {
   render() {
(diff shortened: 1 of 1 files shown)


src/SearchPanel.jsx
```diff
@@ -9,10 +9,10 @@ export class SearchPanel extends React.Component {
   render() {
     var lastUpdatedMessage = null;
 
-    if ( this.props.lastUpdated ) {
+    if ( this.props.lastIndexed ) {
       lastUpdatedMessage =
```
JeffreyMFarley: Probably should change this name too to be consistent

PR #129, opened 2018-02-07
Fix 500 error due to missing user agent

### LINKED ISSUE #128: User-Agent HTTP header is assumed
The code [here](https://github.com/cfpb/ccdb5-ui/blob/master/ccdb5_ui/views.py#L29) assumes that the Django request object always have a `META['HTTP_USER_AGENT']` value, which cannot be assumed. This value gets created from the HTTP request's `User-Agent` header ([see Django docs](https://docs.djangoproject.com/en/1.8/ref/request-response/#django.http.HttpRequest.META)), which may not always exist.

A simple fix here might be to change the code from `request.META['HTTP_USER_AGENT']` to `request.META.get('HTTP_USER_AGENT')`, and handle the case where the value doesn't exist.

## Current behavior

Requests with a search query but no user agent cause a 500 error.

## Expected behavior

Requests with a search query but no user agent are handled properly.

## Steps to replicate behavior (include URLs)

1. Try this curl command, which omits the `User-Agent` header:

```sh
curl -v -H "User-Agent:" https://www.consumerfinance.gov/data-research/consumer-complaints/search/?searchField=company
```

This raises a 500 error.

chosak:
This change fixes the logic that checks the request user agent to see if the browser is unsupported. The current logic assumes that there'll always be an `HTTP_USER_AGENT` header, which is not always the case. The new logic makes this logic more flexible.

This also adds some Python unit tests to verify this behavior and adds instructions on running them to the README.

Fixes #128.

## Additions

- Python unit tests folder.

## Changes

- More flexible logic to support missing `HTTP_USER_AGENT` header.

## Testing

- `./manage.py test`

## Review

- @amymok @JeffreyMFarley @sephcoster 

## Checklist

* [X] Changes are limited to a single goal (no scope creep)
* [X] Code can be automatically merged (no conflicts)
* [X] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [X] Passes all existing automated tests
* [X] New functions include new tests
* [X] New functions are documented (with a description, list of inputs, and expected output)
* [X] Placeholder code is flagged
* [X] Visually tested in supported browsers and devices
* [X] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,13 +110,19 @@ This will push the latest changes to the repo and ensure the new tag is included
 ## How to test the software
 
 #### Unit testing
-To launch the test runner in interactive watch/test mode:
+To launch the JavaScript test runner in interactive watch/test mode:
 
 ```bash
 npm test
 ```
 
 Enter `Control-C` to exit interactive watch mode
+
+To run the Python unit tests:
+
+```bash
+./manage.py test
+```
 
 ## Known issues
 
--- a/ccdb5_ui/tests/test_views.py
+++ b/ccdb5_ui/tests/test_views.py
@@ -1,0 +1,26 @@
+from __future__ import absolute_import, unicode_literals
+
+from django.test import RequestFactory, SimpleTestCase
+
+from ccdb5_ui.views import CCDB5MainView
+
+
+class CCDB5MainViewTest(SimpleTestCase):
+    def setUp(self):
+        self.factory = RequestFactory()
+        self.view = CCDB5MainView.as_view()
+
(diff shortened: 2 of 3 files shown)


chosak:
Thanks for the review @sephcoster. I don't have merge permission for this repo, nor do I have permission to make a new release. I'd like to do both of those things to prevent this 500 from occurring as it (infrequently) does in production. Would you or someone else be able to assist with that?

sephcoster:
✅

PR #134, opened 2018-07-11
Bump Django version support to include 1.11.x

acrewdson:
## Changes

The Django component to this repo has now been tested with 1.11.14, and also remains compatible with Django 1.8.15. We'll assume that means compatibility with versions in between those two.

This change is being made in preparation for upgrading [cfgov-refresh](https://github.com/cfpb/cfgov-refresh) to Django 1.11.

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,13 @@
 ---
 language: node_js
-python: 2.7
+python:
+  - "2.7"
 install:
   - npm install
+  - pip install --user tox
 script:
   - npm test
+  - tox
   - python setup.py bdist_wheel
 after_script:
   - cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js
--- a/README.md
+++ b/README.md
@@ -118,10 +118,11 @@ npm test
 
 Enter `Control-C` to exit interactive watch mode
 
-To run the Python unit tests:
+To run the Python unit tests, first install [Tox](https://tox.readthedocs.io/en/latest/),
+and then run:
 
 ```bash
-./manage.py test
+tox
 ```
 
 ## Known issues
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,14 @@
+[tox]
+skipsdist=True
+
+envlist=dj{18,111}
+
+[testenv]
+install_command=pip install -e ".[testing]" -U {opts} {packages}
+commands=coverage run ./manage.py test
+
+deps=
+    coverage==4.5.1
+
+    dj18: Django>=1.8,<1.9
+    dj111: Django>=1.11,<1.12


acrewdson:
Per @chosak, going to attempt to update the tests to test against multiple Django versions, please hold off on reviewing for now.

README.md
```diff
@@ -24,7 +24,8 @@ Unit testing of the application is performed within
 
 The `ccdb5_ui` (note the underscore) directory contains a thin [Django](https://www.djangoproject.com/)
 implementation that allows it to be used as a plugin for
-[CFPB's public website](https://github.com/cfpb/cfgov-refresh).
+[CFPB's public website](https://github.com/cfpb/cfgov-refresh). To run the Django
+tests, install [Tox](https://tox.readthedocs.io/en/latest/) and then run `tox`.
```
schbetsy: Line 122-ish of this file also has python testing instructions. Do you want to update those to match these?

requirements.txt
```diff
@@ -1 +1 @@
-django==1.8.15
\ No newline at end of file
+django>=1.8.15,<1.12
```
schbetsy: (I'm still learning how python projects work, so forgive me if this is a well-answered question elsewhere). 

[This documentation](https://github.com/cfpb/development/blob/master/standards/python.md#requirementstxt-setuppy-and-specifying-dependencies) specifies that, since we use satellite apps like libraries, we should specify their dependencies in `setup.py` instead of in a `requirements.txt`. 

I know you didn't create this requirements.txt file in the first place, so it wasn't necessarily your choice, but I'm curious: are there pros and cons to the different approaches that we should consider?

README.md
acrewdson: @schbetsy great catch, will fix 👍

requirements.txt
acrewdson: I think it's a great question. In this particular case I'm not too sure what the pros and cons would be in practice, and I don't feel too strongly either way. So in that case, let's see what @chosak thinks. I'm glad you brought this up 👍

.travis.yml
```diff
@@ -1,24 +1,28 @@
...
 install:
-- npm install
+  - npm install
+  - pip install --user tox
 script:
-- npm test
-- python setup.py bdist_wheel
+  - npm test
+  - tox
+  - python setup.py bdist_wheel
```
chosak: Forward looking: use `bdist_wheel --universal` instead [to support Python 3 compatibility](https://wheel.readthedocs.io/en/stable/#defining-the-python-version)?

requirements.txt
chosak: My current preference is to put these dependencies in the `install_requires` in setup.py, and avoid using requirements.txt in satellite apps.

This allows users of this package to do `pip install ccdb5-ui` and have all of the necessary dependencies brought in. As written, the `setup.py` in this package doesn't bring in Django; that means in practice you can't run the tests unless you also install Django manually, or use Tox.

Consider a case where someone is using Django 2.0, and wants to pip install this package. Ideally pip should tell us that it isn't yet compatible with that version of Django, and fail -- that would happen if setup.py contained the constraints.

Using setup.py is nice because if you have a slightly older version of a supported dependency installed first, say, Django 1.11.3, the `install_requires` will match, and your existing version won't be uninstalled/modified. If you don't already have the dependency installed, then the latest compatible version will be installed. Either way you're guaranteed to have a working set of packages.

There are probably some advantages to requirements.txt I'm not thinking of -- for one, it's slightly easier to specify working branches or releases on GitHub (a single line like [this](https://github.com/cfpb/cfgov-refresh/blob/master/requirements/optional-public.txt#L8)) instead of using PyPI installs. [Here's](https://stackoverflow.com/a/20234833) how you have to do it in setup.py.

acrewdson: Thanks @chosak for the excellent thoughts. Would you and @schbetsy suggest I make this change in the present PR? Or open an issue – ?

chosak: My vote would be to include in this PR.

schbetsy: Thanks for the great write-up @chosak . That's super helpful.

acrewdson:
tagging @JeffreyMFarley for a review – thanks!

PR #137, opened 2018-09-21
Properly fix frontend build and wheel names

chosak:
This PR is a followup to #136 which did not correctly ensure that wheel names will be cleanly made.

The frontend build script run as part of `pip install` should do both `npm install` and run the build, so that someone running it locally from scratch will get a working build.

Additionally, on Travis, now that the build is run as part of tox (which runs `pip install`), the package-lock.json file could be modified later in the process. So the reset of that file needs to be moved down to just before the wheel is built.

I ran this against [a release on my fork](https://github.com/chosak/ccdb5-ui/releases/tag/v1.0.13) and the Travis build [created a correctly named wheel](https://travis-ci.org/chosak/ccdb5-ui/builds/431640317):

> creating '/home/travis/build/chosak/ccdb5-ui/dist/ccdb5_ui-1.0.13-py2-none-any.whl' and adding '.' to it

(Travis being unable to upload this to GitHub is an expected failure on my fork that won't happen on this repo.)

I tested this locally by generating a wheel and `pip install`ing it into cfgov-refresh, and can confirm that the minified frontend assets are present even if I have no node_modules present locally. 

To verify wheel name behavior with this branch, you can check it out locally, then create a tag, and then run `python setup.py bdist_wheel` to confirm that the wheel name gets made cleanly.

To verify assets with this branch, you can install it into your cfgov-refresh:

```
$ pip install git+https://github.com/chosak/ccdb5-ui.git@really-fix-wheel-names
```

You should have the minified assets in your virtualenv:

```
$ ls ~/.virtualenvs/cfgov-refresh/lib/python2.7/site-packages/ccdb5_ui/static/
ccdb5.min.js     ccdb5.min.js.map main.css         main.css.map
```

Running a local server you should have a nice-looking page with no missing assets at http://localhost:8000/data-research/consumer-complaints/search/.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,11 +82,6 @@ You will also see any lint errors in the console.
 
 Enter `Control-C` to exit development mode
 
-#### Build deployment package
-
-Our [Travis](https://travis-ci.org/cfpb/ccdb5-ui) configuration is set up to
-build a deployment package after every push to `master`.
-
 ## How to test the software
 
 #### Unit testing


schbetsy:
Looks like the README is now minorly out of date. I don't want to expand the scope of this PR too much, but would you mind updating it?

chosak:
@schbetsy thanks for reviewing. I've removed that section in 8cee66c.

PR #143, opened 2019-06-18
Update cf-icons and cf-forms to version 10 from 4.1.0

anselmbradford:
## Additions
- Add approved.svg and down.svg icons to src directory, for supporting background-image used in cf-forms.

## Changes

- Update cf-icons to 10.1.0.
- Update cf-forms to 10.2.0.
- Convert webfont icon references to SVG.
- Remove webfont icon files from build process and add SVG icon step to build process.
- Update `@cf-icon-path` to utilize path implemented in https://github.com/cfpb/capital-framework/pull/934
- Remove review section from PR template.

## Testing

- Follow https://github.com/cfpb/ccdb5-ui/blob/master/INSTALL.md instructions and ensure icons look similar to those on prod.
- `npm test` should pass.


src/ComplaintDetail.jsx
```diff
@@ -75,7 +75,8 @@ export class ComplaintDetail extends React.Component {
     return (
       <button className="a-btn a-btn__link"
               onClick={this.props.onClickedBack}>
-          <span className="cf-icon cf-icon-left"></span>
+          {/* Includes custom cf-icon-left class to set color. */}
```
wpears: It's probably worth factoring out this svg into it's own function since it's duplicated below

src/ComplaintDetail.jsx
```diff
@@ -86,46 +87,62 @@ export class ComplaintDetail extends React.Component {
...
+        </span>
         <span className="body-copy">{ value }</span>
       </div>
     )
   }
 
   _renderConsumerConsent( value ) {
+    // Each SVG includes custom classes for setting color.
```
wpears: The iconMap should probably be generated in the constructor (like the renderMap) so the svg elements are only created once and reused on further renders.

PR #159, opened 2019-12-12
View reducer

flacoman91:
reducer for view model, 
added actions for date intervals
updates to date filter

### DIFF
--- a/src/actions/filter.jsx
+++ b/src/actions/filter.jsx
@@ -1,4 +1,4 @@
-import { REQUERY_ALWAYS, REQUERY_HITS_ONLY } from '../constants'
+import { REQUERY_ALWAYS } from '../constants'
 
 export const DATE_INTERVAL_CHANGED = 'DATE_INTERVAL_CHANGED'
 export const DATE_RANGE_CHANGED = 'DATE_RANGE_CHANGED'
--- a/src/actions/__tests__/filter.spec.jsx
+++ b/src/actions/__tests__/filter.spec.jsx
@@ -1,4 +1,4 @@
-import { REQUERY_ALWAYS, REQUERY_HITS_ONLY } from '../../constants'
+import { REQUERY_ALWAYS } from '../../constants'
 import * as sut from '../filter'
 
 describe('action:filterActions', () => {
@@ -12,7 +12,7 @@ describe('action:filterActions', () => {
           filterName,
           minDate,
           maxDate,
-          requery: REQUERY_HITS_ONLY
+          requery: REQUERY_ALWAYS
         }
         expect(sut.changeDateRange(filterName, minDate, maxDate))
           .toEqual( expectedAction );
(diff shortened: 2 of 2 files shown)


src/__tests__/App.spec.jsx
```diff
@@ -9,6 +9,14 @@ import thunk from 'redux-thunk'
...
   it('renders without crashing', () => {
+
+    // set the date so snapshot will always be the same.
+    const DATE_TO_USE = new Date('2016');
+    const _Date = Date;
+    global.Date = jest.fn(() => DATE_TO_USE);
+    global.Date.UTC = _Date.UTC;
+    global.Date.parse = _Date.parse;
+    global.Date.now = _Date.now;
     defaultQuery.searchText = 'foo';
```
JeffreyMFarley: Good job.  I'm used to Jasmine having dates as part of its mocking system

src/actions/filter.jsx
```diff
@@ -29,6 +30,20 @@ export function changeDateRange( filterName, minDate, maxDate ) {
...
+ * Notifies the application that date interval (m, w, yr) was toggled
+ *
+ * @param {string} dateInterval which filter was clicked
+ * @returns {string} a packaged payload to be used by Redux reducers
+ */
+export function dateIntervalToggled( dateInterval ) {
+  return {
+    type: DATE_INTERVAL_CHANGED,
+    dateInterval,
+    requery: REQUERY_HITS_ONLY
```
JeffreyMFarley: For a date __filter__ on the left-side this should be `REQUERY_BOTH` if we are using two queries.

src/actions/view.jsx
```diff
@@ -0,0 +1,20 @@
...
+ * Notifies the application that the tab has changed
+ *
+ * @param {string} tab the new tab name
+ * @returns {string} a packaged payload to be used by Redux reducers
+ */
+export function tabChanged( tab ) {
+  return {
+    type: TAB_CHANGED,
+    tab,
+    requery: REQUERY_HITS_ONLY
```
JeffreyMFarley: This is definitely a right-hand only query

src/reducers/__tests__/query.spec.jsx
```diff
@@ -5,18 +5,29 @@ import actions from '../../actions'
...
+        })
+      // doing this because I can't seem to mock the date since
+      // defaultQuery is imported
+      expect( res ).toHaveProperty( 'date_received_max' )
+      expect( res ).toHaveProperty( 'date_received_min' )
+      expect( res.queryString ).toContain( 'date_received_max' )
+      expect( res.queryString ).toContain( 'date_received_min' )
+      expect( res.queryString ).toContain( 'field=all&page=1&size=25' +
+        '&sort=created_date_desc' )
+    })
```
JeffreyMFarley: Yeah, the way it is set up in the reducer, you would need to mock the date before all the imports are read.  That would mean breaking out `proxyquire` :shudder:

This is perfectly fine

src/reducers/query.jsx
```diff
@@ -108,6 +112,38 @@ function processParams( state, action ) {
...
+  if ( res[dateInterval] ) {
+    newState.date_received_min = res[dateInterval]
+  } else if ( dateInterval === 'All' ) {
+    newState.date_received_min = new Date( types.DATE_RANGE_MIN )
+  }
+
+  newState.date_received_max = new Date()
+
+  return newState;
+}
```
JeffreyMFarley: :+1:

PR #162, opened 2019-12-19
Tile map component

flacoman91:
Tile map component

### DIFF
--- a/src/__tests__/utils.spec.jsx
+++ b/src/__tests__/utils.spec.jsx
@@ -1,4 +1,4 @@
-import { clamp, coalesce, debounce, shortIsoFormat } from '../utils'
+import { clamp, coalesce, debounce, hashCode, shortIsoFormat } from '../utils'
 import React from 'react'
 
 describe('module::utils', () => {
@@ -66,6 +66,15 @@ describe('module::utils', () => {
 
       actual = clamp( 10, 15, 25 );
       expect( actual ).toEqual( 15 );
+    } );
+  })
+
+  describe('hashCode', ()=>{
+    it( 'hashes strings', () => {
+      let actual = hashCode( '' );
+      expect( actual ).toEqual( 0 );
+      actual = hashCode( 'foobar' );
+      expect( actual ).toEqual( -1268878963 );
     } );
   })
 })


config/jest/fileTransform.js
```diff
@@ -1,12 +1,40 @@
...
+            key: null,
+            props: Object.assign({}, props, {
+              children: ${assetFilename}
+            })
+          };
+        }),
+      };`;
+    }
+
+    return `module.exports = ${assetFilename};`;
```
JeffreyMFarley: This block of code seems important.  What does it do?

src/TileChartMap.jsx
```diff
@@ -0,0 +1,54 @@
...
+export class TileChartMap extends React.Component {
+  componentDidUpdate( prevProps ) {
+    const props = this.props
+    if ( !props.data[0].length ) {
+      return
+    }
+
+    if ( JSON.stringify( prevProps.data ) !== JSON.stringify( props.data ) ) {
+      this._redrawMap()
+    }
```
JeffreyMFarley: Depending on the size of the data, this could be expensive for a difference check.  Do we have the hashing functions in CCDB?

src/TileChartMap.jsx
```diff
@@ -0,0 +1,54 @@
...
+    ]
+
+    // eslint-disable-next-line no-unused-vars
+    const chart = new TileMap( {
+      el: document.getElementById( 'tile-chart-map' ),
+      data: this.props.data,
+      colors,
+      localize: true
+    } )
+  }
```
JeffreyMFarley: I'm assuming `chart` is not used because `new TileMap` embeds itself into the DOM at `document.getElementById( 'tile-chart-map' )`

Assigning it to `const _` might get rid of the warning

src/__tests__/TileChartMap.spec.jsx
```diff
@@ -0,0 +1,97 @@
...
+import { TileMap } from 'cfpb-chart-builder'
+
+// this is how you override and mock an imported constructor
+jest.mock( 'cfpb-chart-builder', () => {
+  return {
+    TileMap: jest.fn().mockImplementation( () => {
+      return {}
+    } )
+  }
+} )
```
JeffreyMFarley: Very cool!

src/__tests__/TileChartMap.spec.jsx
```diff
@@ -0,0 +1,97 @@
...
+    } )
+  } )
+
+  describe( 'componentDidUpdate', () => {
+    beforeAll( () => {
+      const div = document.createElement( 'div' )
+      div.setAttribute( 'id', 'tile-chart-map' )
+      window.domNode = div
+      document.body.appendChild( div )
+    } )
```
JeffreyMFarley: nice

config/jest/fileTransform.js
flacoman91: It came with the updated ejected Create React App.  with the updated jest config files, it wasn't working if I didn't include it.

src/TileChartMap.jsx
flacoman91: No, doesn't look like we have a hashing function.  I can port it over from explorer.

src/TileChartMap.jsx
flacoman91: I tried that and it's still nagging me about the error.    Chart isn't used, it's just how the constructor works.   I can remove the const chart =  and do this instead

https://stackoverflow.com/questions/33127663/pattern-for-javascript-instantiation-which-doesnt-upset-eslint

But I'm throwing in another eslint ignore message

JeffreyMFarley: It's fine as is, these edge cases are annoying to try and stay compliant with

src/TileChartMap.jsx
JeffreyMFarley: Ok, the underlying code still uses `JSON.stringify`, but would could store the resulting hash in the properties in the future if `.data` becomes too large

flacoman91: haha, i literally just moved this over and saw the same thing.  it is a bit cleaner and JSON.stringify is being used all over the place in the componentDidUpdate handler.

I think what the hash function does better is that you have a shorter value stored in memory vs a super long string.  We should convert all of the JSON.stringify to use the hash in a separate PR in one PR in another chore.

JeffreyMFarley: Agreed!

JeffreyMFarley: Agreed

PR #163, opened 2019-12-20
Map tab

flacoman91:
adding a map tab and also adding a sweet tile map onto that.


src/MapPanel.jsx
```diff
@@ -0,0 +1,30 @@
...
+import DateIntervals from './DateIntervals';
+import Loading from './Dialogs/Loading'
+import { MemoryRouter } from 'react-router'
+import React from 'react'
+import TileChartMap from './TileChartMap';
+
+export class MapPanel extends React.Component {
+  render() {
+    return (
+      <MemoryRouter>
```
sephcoster: Using MemoryRouter here could prevent users from using their back / forward buttons and typically also does not update the application URL. Is that the intent for our the tab structure, or is this here mostly for testing / we plan to move this to a more conventional router at some point in the browser?

flacoman91: short answer, I don't know.  I used the original results panel as a baseline boilerplate and put put things in it to get it to work.   I thought that List, Trends and Map all have a similar structure and I did not know what Memory Router did

flacoman91: Ok, so it turns out MemoryRouter isn't needed here and was a big red herring.

The broken fwd/back functionality was introduced by new params I added in via the query reducer which caused the history to break.

src/reducers/query.jsx
```diff
@@ -481,6 +516,12 @@ export function stateToQS( state ) {
     }
   } )
 
+  // exclude certain params from the URL since it's a calculated value
+  // coming from db and cant be restored
+  excludeParams.forEach( o => {
+    delete state[o]
+  } )
+
```
JeffreyMFarley: 👍

PR #166, opened 2020-01-03
Row charts / Map Tab

flacoman91:
Map Tab, updated with 2 row charts mapped to products and issues


flacoman91:
> Is this still Draft/WIP?

whoops, sorry!  forgot to toggle it off.  I changed the name though.

PR #173, opened 2020-02-04
Per capita

flacoman91:
Adding per Capita 

per-capita

### DIFF
--- a/src/constants/index.jsx
+++ b/src/constants/index.jsx
@@ -188,6 +188,11 @@ export const TILE_MAP_STATES = [
   'WY'
 ]
 
+// 2017 ACS census via the census API:
+// https://api.census.gov/data/2017/acs/acs5?get=B01003_001E,NAME&for=state
+//
+// B01003_001E is defined as TOTAL POPULATION
+// (detailed here: https://api.census.gov/data/2016/acs/acs5/variables.html)
 export const STATE_DATA = [
   {
     abbr: 'AL',


src/TileChartMap.jsx
```diff
@@ -74,6 +75,43 @@ export class TileChartMap extends React.Component {
...
+  return showDefault ? stateObj.value : getPerCapita( stateObj )
+}
+
+/**
+ * helper function to calculate percapita value
+ * @param {object} stateObj a state containing abbr and value
+ * @returns {string} the per capita value
+ */
+function getPerCapita( stateObj ) {
+  const pop = STATE_DATA.find( o => o.abbr === stateObj.name ).population
```
sephcoster: Any advantage to precomputing this on page load after render so it's available on click, rather than doing the read each time, or is that overoptimization for this?

flacoman91: Whoops. you're right. I updated the logic to have it in the reducer.

src/DateIntervals.jsx
```diff
@@ -11,14 +11,15 @@ export class DateIntervals extends React.Component {
   }
 
   _btnClassName( dateInterval ) {
-    const btnClass = 'date-selector ';
+    const btnClass = 'date-selector interval-' + dateInterval
     return dateInterval === this.props.dateInterval ?
-      btnClass + 'selected' : btnClass;
+      btnClass + ' selected' : btnClass
   }
```
JeffreyMFarley: Without pesky white space adds

```javascript
const classes = ['date-selector', 'interval-' + dateInterval]
if (dateInterval === this.props.dateInterval) {
    classes.push('selected')
}
return classes.join(' ')
```

src/PerCapita.jsx
```diff
@@ -0,0 +1,33 @@
...
+  dataNormalization: state.map.dataNormalization
+} );
+
+export const mapDispatchToProps = dispatch => ( {
+  onDataNormalization: ev => {
+    dispatch( dataNormalizationChanged( ev.target.value ) )
+  }
+} );
+
+export default connect( mapStateToProps, mapDispatchToProps )( PerCapita )
```
JeffreyMFarley: :chefs_kiss:

src/TileChartMap.jsx
```diff
@@ -87,6 +114,7 @@ export const processStates = state => {
   const statesFilter = state.query.state || []
   const states = state.map.state
   const stateData = states.map( o => {
+    o.perCapita = getPerCapita( o )
     o.className = getStateClass( statesFilter, o.name )
```
JeffreyMFarley: I like this pipeline-y approach to enhancing the state data

src/actions/map.jsx
```diff
@@ -1,9 +1,24 @@
...
+
+/**
+ * Indicates data normalization (yes, per capita) value was selected
+ * @param {object} value of the data normalization
+ * @returns {object} a packaged payload to be used by Redux reducers
+ */
+export function dataNormalizationChanged( value ) {
+  return {
+    type: DATA_NORMALIZATION_SELECTED,
+    requery: REQUERY_NEVER,
```
JeffreyMFarley: :+1:

src/constants/index.jsx
```diff
@@ -183,3 +187,318 @@ export const TILE_MAP_STATES = [
   'WV',
   'WY'
 ]
+
+export const STATE_DATA = [
```
JeffreyMFarley: This is a nitpick, but we may want to put the source of the data in a comment above this block so future maintainers (us) will know how to update.

src/reducers/__tests__/map.spec.jsx
```diff
@@ -243,6 +257,29 @@ describe( 'reducer:map', () => {
...
+
+    it( 'handles empty params', () => {
+      expect( target( state, action ) ).toEqual( state )
+    } )
+
+    it( 'handles dataNormalization params', () => {
+      action.params = { dataNormalization: 'hello' }
+      const actual = target( state, action )
+      expect( actual.dataNormalization ).toEqual( 'hello' )
+    } )
```
JeffreyMFarley: :+1:

src/DateIntervals.jsx
flacoman91: thank you!

src/constants/index.jsx
flacoman91: Added, from admiral's comment here:
https://github.cfpb.gov/data-products-team/explorer-docs/issues/849#issuecomment-236310

PR #178, opened 2020-02-13
fixing styling issues

flacoman91:
adding more padding, other various fixes

### DIFF
--- a/src/RowChart.jsx
+++ b/src/RowChart.jsx
@@ -106,7 +106,7 @@ export class RowChart extends React.Component {
 
   render() {
     return (
-      <div>
+      <div class="product-section">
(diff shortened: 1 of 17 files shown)


src/RowChart.jsx
```diff
@@ -106,7 +106,7 @@ export class RowChart extends React.Component {
 
   render() {
     return (
-      <div>
+      <div class="product-section">
```
flacoman91: this is a generic row chart, so it could be either issue or product, or anything else.

also in the component syntax this should be className
I'd change this to:
className="row-chart-section"

flacoman91:
closing this since i am updating your branch and working off of that.

PR #190, opened 2020-03-09
fixing row chart and map in IE11

flacoman91:
Fixing busted object properties
Adding error checking for the dom.node.children methods


flacoman91:
Thanks. I'll merge this and continue to fix the styling for the map in ie11 since it looks like this currently:

Screen Shot 2020-03-10 at 9 40 43 AM

PR #188, opened 2020-03-04
Make charts and maps work better at different sizes

JeffreyMFarley:
The application will be used on a vast array of devices, so it is important that they look good at all sizes. 

## Testing

- 💯 

## Screenshots
ccdb5-sizes

## Notes

- The map and list have different breakpoints for hiding the filters
    - Map at 750px
    - List at 599px

## Todos

- Test on older IE versions

## Checklist

* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [x] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [x] Passes all existing automated tests
* [x] New functions include new tests
* [x] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #191, opened 2020-03-11
Fixing the map display in IE 11

flacoman91:
IE11 support of flex box is not good, so it must be fixed.

No breaking changes with Chrome:
Chrome

IE 11 fixes, using the breakpoints in the original responsive [ticket](https://github.com/cfpb/ccdb5-ui/pull/188)
widths: 1200, 901, 899, 749, 599, 375
IE11

### DIFF
--- a/src/TileMap/index.js
+++ b/src/TileMap/index.js
@@ -174,8 +174,6 @@ export function tileFormatter() {
     iePatch = '<br />'
   }
 
-  // const iePatch = !!window.MSInputMethodContext && !!document.documentMode ?
-  //   '<br />' : '';
   const value = this.point.displayValue.toLocaleString();
   return '<div class="highcharts-data-label-state">' +
     '<span class="abbr">' + this.point.name + '</span>' +


JeffreyMFarley:
Unfortunately, here is the matrix for checking the views

Browser|1200 px|1000 px|751 px|749 px|599 px|< 400 px
---|---|---|---|---|---|---
Chrome|||Filters|No filters||No Values
IE 11|

flacoman91:
> Unfortunately, here is the matrix for checking the views
> 
> Browser	1200 px	1000 px	751 px	749 px	599 px	< 400 px
> Chrome			Filters	No filters		No Values
> IE 11

back to the drawing board for me then.

flacoman91:
Because pics or it didn't happen:

IE11-751
IE11-1200
IE11-1000

IE11-399
IE11-599
IE11-749

flacoman91:
Chrome 
Chrome

src/TileChartMap.less
```diff
@@ -146,10 +146,6 @@ x/y positioning on the SVG elements.
         .legend-box {
             stroke: @gray;
         }
-
-        .legend-text {
-            transform: translate(3px, -3px);
-        }
```
JeffreyMFarley: In the screenshots, the tile text looks shifted.  Do we need to restore this?

src/TileChartMap.less
```diff
@@ -249,6 +245,11 @@ x/y positioning on the SVG elements.
         display: flex;
         flex-direction: row;
         flex-wrap: wrap;
+        // IE 11 fixes
+        @media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
+            display: inline-block;
+            float: left;
+        }
```
JeffreyMFarley: :+1:

src/TileMap/index.js
```diff
@@ -169,9 +169,17 @@ export function getColorByValue( value, bins ) {
  * @returns {string} html output
  */
 export function tileFormatter() {
+  let iePatch = ''
+  if ( navigator.userAgent.indexOf( 'MSIE' ) !== -1 || navigator.appVersion.indexOf( 'Trident/' ) > -1 ) {
+    iePatch = '<br />'
+  }
```
JeffreyMFarley: :+1:

src/TileChartMap.less
flacoman91: this is only for the legend bin values, and it was moved in the legend code here https://github.com/cfpb/ccdb5-ui/pull/191/files#diff-a7f3f10a59c833f0a7e2a676fcff76f4R288

For some reason IE11 wasn't respecting this css rule

PR #193, opened 2020-03-17
Provide a standard error block

JeffreyMFarley:
It was reported back on February 14th that the application was not handling errors correctly on the map page.  This is no longer the case, but it should probably be more specific when an error occurs

## Testing

- :100:

#### List Tab

list error

#### Map Tab

map error

## Checklist

### DIFF
--- a/src/MapPanel.jsx
+++ b/src/MapPanel.jsx
@@ -15,9 +15,8 @@ export class MapPanel extends React.Component {
     return (
       <section className="map-panel">
         <ActionBar/>
-        { this.props.error ?
-          <ErrorBlock text="There was a problem executing your search" /> :
-          null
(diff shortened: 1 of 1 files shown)


src/MapPanel.jsx
```diff
@@ -14,6 +15,10 @@ export class MapPanel extends React.Component {
     return (
       <section className="map-panel">
         <ActionBar/>
+        { this.props.error ?
+          <ErrorBlock text="There was a problem executing your search" /> :
```
flacoman91: this works, but you can also do this to avoid that ternary

```
{ this.props.error &&
          <ErrorBlock text="There was a problem executing your search" />
}
```

PR #201, opened 2020-03-20
Accessibility Updates to CCDB

JeffreyMFarley:
This PR fixes a few results of an accessibility check using the [WAVE evaluation tool](https://wave.webaim.org/)

- [x] 10X Missing Form Labels in Filter Directives (all dates, all type-aheads)
- [x] Empty form label on main search bar (A form label is present, but does not contain any content)
- [x] 4X Broken ARIA references on our Date Input fields
- [x] 4X Orphaned Form labels on Date Input Fields
- [x] 3X Headings Skipped - (Beta, Search Within, Exports) - is this on purpose because we use the styles but they are not headings we want screen readers to capture?
- [x] 6X Redundant Title Text is the same as the Alt text. Remove Title text in this case according to WAVE - this mostly occurs for filter values in the Product and Issue filter

## Notes

We will get the empty button error for each pill because of how it is constructed

```Javascript
    <li className="pill flex-fixed">
      <span className="name">{ trimmed }</span>
      <button onClick={ remove }
              title={'Remove ' + trimmed + ' as a filter'}>
        { iconMap.getIcon( 'delete' ) }
      </button>
    </li>
```

The good news is it works correctly for screen readers and does handle the accessible navigation correctly.  So we will just have to ignore these WAVE errors

## Testing

- :100:

## Checklist

* [x] Changes are limited to a single goal (no scope creep)
* [x] Code can be automatically merged (no conflicts)
* [x] Code follows the standards laid out in the [front end playbook](https://github.com/cfpb/front-end)
* [x] Passes all existing automated tests
* [x] New functions include new tests
* [x] New functions are documented (with a description, list of inputs, and expected output)
* [ ] Placeholder code is flagged
* [ ] Visually tested in supported browsers and devices
* [ ] Project documentation has been updated (including the "Unreleased" section of the CHANGELOG)

PR #203, opened 2020-03-24
Improve the experience for screen readers (plus a few other changes)

JeffreyMFarley:
Similar to #201, this PR contains improvements for accessibility. Specifically [screen reader support](https://help.apple.com/voiceover/mac/10.15/#/vo2682).

## Pills are updated to help with UX, WAVE and Screen Readers

#201 had a hand-wavy fix for WAVE, this PR contains a compliant fix while maintaining the required UX that only the `x` is considered the button

##  Tab navigation should go from the search bar to the count for the search, then to the explicit results, instead of to the filters

Added a skip to results link that is appears when focused via keyboard

skip-to

## Add more alt / title text to Complaint number headers so screen readers know what the number means, not just what it says

Based on [an article](https://thatdevgirl.com/blog/accessibility-phone-number-formatting), the `aria-label` for a complaint number is read out one digit at a time

complaint-readout

## Read out each field of a complaint or complaint detail

Added `tabIndex="0"` so that each piece of information of a complaint card can be reached via keyboard and read by the screen reader

complaint-card

## Fixed the missing check mark

Screen Shot 2020-03-24 at 9 54 06 AM

### DIFF
--- a/src/utils.jsx
+++ b/src/utils.jsx
@@ -9,7 +9,7 @@ import moment from 'moment'
 * @returns {string} an expanded string of digits
 */
 export function ariaReadoutNumbers( s ) {
-  return Array.from( s ).join( ' ' )
+  return Array.from( s || '' ).join( ' ' )
 }
 
 // eslint-disable-next-line complexity
--- a/src/__tests__/utils.spec.jsx
+++ b/src/__tests__/utils.spec.jsx
@@ -1,12 +1,29 @@
 import {
-  calculateDateInterval, clamp, coalesce, debounce, getFullUrl, hashCode,
-  shortIsoFormat
+  ariaReadoutNumbers, calculateDateInterval, clamp, coalesce, debounce,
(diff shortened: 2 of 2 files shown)


src/utils.jsx
```diff
@@ -1,6 +1,17 @@
...
 import moment from 'moment'
 
+/**
+* Breaks up '123' to '1 2 3' to help screen readers read digits individually
+* https://thatdevgirl.com/blog/accessibility-phone-number-formatting
+*
+* @param {string} s the string of digits
+* @returns {string} an expanded string of digits
+*/
+export function ariaReadoutNumbers( s ) {
```
flacoman91: slick 👍

flacoman91: I know this is tested via the snapshots, but can we get a test?

JeffreyMFarley: sure!

PR #206, opened 2020-03-26
Conditionally disable complaints per 1000

flacoman91:
When any filters are applied other than Date,
- [x] Switch Map Shading dropdown to 2 button treatment on Landing Page
- [x] Disable the Complaints Per 1000 Button
- [x] Go back to raw Complaints Mode
- [x] Display warning message

filter-apply

### DIFF
--- a/src/__tests__/utils.spec.jsx
+++ b/src/__tests__/utils.spec.jsx
@@ -181,6 +181,17 @@ describe('module::utils', () => {
 
       expect( hasFiltersEnabled( query ) ).toBeTruthy()
     } )
+
+    it( 'handles company_received filters', () => {
+      const query = {
+        date: {},
+        bogus: {},
+        product: [],
+        company_received_max: 'foo'
+      }
+
+      expect( hasFiltersEnabled( query ) ).toBeTruthy()
+    } )
   } )
 })
 
--- a/src/__tests__/__snapshots__/App.spec.jsx.snap
+++ b/src/__tests__/__snapshots__/App.spec.jsx.snap
@@ -1233,12 +1233,6 @@ exports[`initial state renders without crashing 1`] = `
               <p>
                 Map Shading
               </p>
-              <label
-                className="u-visually-hidden"
-                htmlFor="data-normalization"
-              >
-                Select the kind of shading to apply to the map
-              </label>
               <button
                 className="a-btn raw "
                 onClick={[Function]}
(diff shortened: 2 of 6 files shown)


src/PerCapita.less
```diff
@@ -1,7 +1,10 @@
 @import (less) "base.less";
 
 .per-capita {
-  select {
-    height: 34px;
+  button {
+    color: @black;
+    &.deselected {
+      background-color: @pacific-40;
```
sephcoster: It looks like this color may be slightly off compared to the date range selector:
Screen Shot 2020-03-26 at 1 09 50 PM

The contrast difference with black text is subtle but makes it difficult to read.

Minor fix if you can get to it in this PR. Thanks!

src/__tests__/utils.spec.jsx
```diff
@@ -149,5 +150,27 @@ describe('module::utils', () => {
...
+        product: []
+      }
+
+      expect( hasFiltersEnabled( query ) ).toBeFalsy()
+    } )
+
+    it( 'handles some filters', () => {
+      const query = {
+        date: {},
+        bogus: {},
```
JeffreyMFarley: :+1:

flacoman91:
Ok, updated colors, and added the narrative and fixed the tags filter.
filter-apply-tags
filter-apply-2

the filters that trigger this are:
```
  'company',
  'company_public_response',
  'company_response',
  'consumer_consent_provided',
  'consumer_disputed',
  'issue',
  'product',
  'state',
  'submitted_via',
  'tags',
  'timely',
  'zip_code',

'has_narrative'
```

still up in the air about
```
  'company_received_max',
  'company_received_min',
```

src/PerCapita.jsx
```diff
@@ -2,35 +2,58 @@ import './PerCapita.less'
...
-                onChange={this.props.onDataNormalization}
-                id="data-normalization">
-          <option value={GEO_NORM_NONE}>Complaints</option>
-          <option value={GEO_NORM_PER1000}>Per 1000 population</option>
-        </select>
+        <div className="per-capita m-btn-group">
+          <p>Map Shading</p>
+          <label className="u-visually-hidden" htmlFor="data-normalization">
+            Select the kind of shading to apply to the map
+          </label>
```
JeffreyMFarley: You should get rid of this since the drop-down disappeared.  It triggers a new WAVE warning

flacoman91: thanks for validating.  Will 🔪 this

sephcoster:
These changes look good. company_received (both) should also be on the trigger list. Thanks.

src/PerCapita.less
flacoman91: changed.

PR #209, opened 2020-03-27
Ie11 missing polyfills

flacoman91:
Sorry,  missed another polyfill. 

The tile map was rendering, but the values were 0...

it said .values was undefined.

the js-doc ignore is just to get around the eslint nag since we clearly did define the comments.

Screen Shot 2020-03-26 at 8 25 18 PM

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,5 @@
 import 'core-js/es6/array'
+import 'core-js/es7/object'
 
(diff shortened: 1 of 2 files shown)


flacoman91:
> Thanks for being on top of this

and thanks for being on top of WAVE accessibility issues 👍

PR #211, opened 2020-03-27
Mobile Filters

adulfan:
Make filters visible at mobile width for both Map and List view

- [x] remove line under Showing N of NNN complaints
- [x] add buttons to toggle open / close filters
- [x] add interactivity
- [x] update snapshots
- [x] update tests

Map View:
- [x] Standardize button toggles for Date Interval and Per Capita

List View:
- [x] Update styling of the Sort and Show NN results dropdown

## Screenshots
**Map View:**
map-view-responsive

**List View**
list-view-responsive

### DIFF
--- a/src/ListPanel.jsx
+++ b/src/ListPanel.jsx
@@ -53,7 +53,7 @@ export class ListPanel extends React.Component {
           <FilterPanelToggle/>
           <Separator />
           <section className="cf-select">
-            <label className="u-visually-hidden" htmlFor="choose-sort">
+            <label className="u-visually-hidden" htmlFor="choose-size">
               Select the number of results to display at a time
             </label>
             <p>
--- a/src/__tests__/__snapshots__/ListPanel.spec.jsx.snap
+++ b/src/__tests__/__snapshots__/ListPanel.spec.jsx.snap
@@ -68,7 +68,7 @@ exports[`component:ListPanel displays a message when an error has occurred 1`] =
     >
       <label
         className="u-visually-hidden"
-        htmlFor="choose-sort"
+        htmlFor="choose-size"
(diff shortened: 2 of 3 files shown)


JeffreyMFarley:
Interesting!  For some reason Anna's fork is not triggering Travis builds.

When I run `npm test`, I get 15 failing snapshot tests

sephcoster:
It looks like the account associated with the pushes on her local machine may not be associated with github.com? @adulfan you may want to look into that. Could be commits from non-github.com accounts don't run the builds?

src/MapPanel.jsx
```diff
@@ -22,6 +22,7 @@ const WARNING_MESSAGE = 'Due to your filter selections, the “Complaints per' +
 
 export class MapPanel extends React.Component {
   render() {
+    /* eslint complexity: ["error", 5] */
```
JeffreyMFarley: These types of comment apply to the whole module, no the next line... so I just move them to the top of the file

`// eslint-disable-next-line complexity`

is what to use for a single line function that is ridiculous

src/ListPanel.jsx
```diff
@@ -31,6 +48,39 @@ export class ListPanel extends React.Component {
       <section className="list-panel">
         <ActionBar />
         <StaleDataWarnings />
+        { this.props.showMobileFilters && <FilterPanel/> }
+        <div className="layout-row refine">
+          <FilterPanelToggle/>
+          <Separator />
+          <section className="cf-select">
+            <label className="u-visually-hidden" htmlFor="choose-sort">
```
JeffreyMFarley, line 56: should be `htmlFor=choose-size`

PR #221, opened 2020-04-03
Dismiss Complaints per 1,000 warning

flacoman91:
Add the ability to dismiss the Per Capita Map Warning

Logic of when the warning is displayed and interaction: 

Default: 
No message

Toggle Filter ->
  - Show Warning
  - Disable Complaints per 1,000 button
  - User dismisses Message 
    - Disable Message

Toggle Filter ->
  - Nothing Happens

Remove a filter
  - If filters still applied
    - do nothing
  - else if no filters applied
    - Enable Complaints per 1,000 button
    - Reenable Warning Message, but do not display it

dismissnag

Screen Shot 2020-04-03 at 4 07 58 PM
Screen Shot 2020-04-03 at 4 07 46 PM

### DIFF
--- a/src/reducers/query.jsx
+++ b/src/reducers/query.jsx
@@ -189,8 +189,6 @@ function processParams( state, action ) {
     }
   } )
 
-  validatePer1000( processed )
-
   return alignIntervalAndRange( processed )
 }
 
@@ -297,8 +295,6 @@ export function toggleFlagFilter( state, action ) {
     }
   } )
 
-  validatePer1000( newState )
-
   return newState
 }
 
@@ -339,7 +335,6 @@ export function addMultipleFilters( state, action ) {
 
   newState[name] = a
 
-  validatePer1000( newState )
   return newState
 }
 
@@ -377,8 +372,6 @@ export function toggleFilter( state, action ) {
     )
   }
 
-  validatePer1000( newState )
-
   return newState
 }
 
@@ -401,8 +394,6 @@ export function addStateFilter( state, action ) {
     state: stateFilters
   }
 
-  validatePer1000( newState )
-
   return newState
 }
 
@@ -417,8 +408,6 @@ export function clearStateFilter( state ) {
     ...state,
     state: []
   }
-
-  validatePer1000( newState )
 
   return newState
 }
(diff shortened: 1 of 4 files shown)


flacoman91:
hold off til Matt provides a mock of what the close button should look like

flacoman91:
Updated screenshots of the close button:
Screen Shot 2020-04-06 at 6 01 55 PM
Screen Shot 2020-04-06 at 6 02 13 PM

src/Warning.less
```diff
@@ -0,0 +1,13 @@
+@import (less) "./base.less";
+
+.warning {
+    .close {
+        position: absolute;
+        right: 0;
+        top: 0;
+        margin: 5px 10px;
+        svg {
+            fill: @pacific;
```
JeffreyMFarley, line 10: Blue and not black? Interesting

src/__tests__/Warning.spec.jsx
```diff
@@ -0,0 +1,36 @@
...
+    const wrapper = shallow( <Warning { ...props }/> )
+    expect( wrapper.find('.m-notification_message')
+      .equals(<div className="h4 m-notification_message">Some nag message you can close</div>) )
+      .toEqual( true )
+
+    expect( wrapper.find('svg').length ).toEqual( 2 )
+    
+    const button = wrapper.find( '.close' )
+    button.simulate( 'click' )
+    expect( props.closeFn ).toHaveBeenCalled()
```
JeffreyMFarley, line 33: 🥇

src/reducers/query.jsx
```diff
@@ -187,6 +189,8 @@ function processParams( state, action ) {
     }
   } )
 
+  validatePer1000( processed )
```
JeffreyMFarley, line 192: These could probably all move to above Line 778 (making the query string) since it is a calculated state

src/Warning.less
flacoman91: I confirmed with Matt. he said interactive color #0072CE since its a clickable link

src/reducers/query.jsx
flacoman91: You're right. I just tested it.  I had a hunch that I was putting the same line in so many places, but thought it was dependent only on a few actions.  I will change this.

Thank you

PR #224, opened 2020-04-06
Row chart cut off numbers

flacoman91:
Fixing the styling of the chart to show cut off numbers

bonus:
updated cfpb britecharts dependency to the latess
tweaked width of the row chart at desktop widths to make the charts full width

fix-cut-off-numbers-2
fix-cut-off-numbers


flacoman91:
sorry i need to update test coverage

PR #232, opened 2020-04-15
Remove window watchers in Row & Map chart 

flacoman91:
Removing window watchers since we don't need them anymore since we keep track of width in redux now

Test coverage: 100%
 
Works like before

### DIFF
--- a/src/RowChart.jsx
+++ b/src/RowChart.jsx
@@ -1,7 +1,6 @@
 import './RowChart.less'
 import * as d3 from 'd3'
 import { connect } from 'react-redux'
-import { debounce } from './utils'
 import { max } from 'd3-array'
 import React from 'react'
 import { row } from 'britecharts'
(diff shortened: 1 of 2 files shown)


src/RowChart.jsx
```diff
@@ -64,18 +64,11 @@ export class RowChart extends React.Component {
...
-  componentDidMount() {
-    window.addEventListener( 'resize', this._throttledRedraw );
-  }
 
   componentDidUpdate() {
     this._redrawChart()
   }
 
-  componentWillUnmount() {
-    window.removeEventListener( 'resize', this._throttledRedraw );
```
JeffreyMFarley, line 76: Can we get rid of `this._throttledRedraw` in the constructor too?

src/TileChartMap.jsx
```diff
@@ -14,10 +14,6 @@ export class TileChartMap extends React.Component {
     this._throttledRedraw = debounce( this._redrawMap.bind( this ), 200 );
   }
 
-  componentDidMount() {
-    window.addEventListener( 'resize', this._throttledRedraw );
```
JeffreyMFarley, line 18: Same comment here about `this._throttledRedraw`

src/RowChart.jsx
flacoman91: 💥  gone 👍

PR #234, opened 2020-04-16
Print friendly Page

flacoman91:
Print friendly page for CCDB-UI

Clicking Print-friendly view opens a a print friendly formatted page in a new tab

Ctrl/Command+P still works and the end result is the same.

print-friendly
print-friendly-2

Testing: 100%

### DIFF
--- a/src/WindowSize.jsx
+++ b/src/WindowSize.jsx
@@ -21,15 +21,11 @@ export class WindowSize extends React.Component {
   }
 }
 
-const mapStateToProps = state => ( {
-  printMode: state.view.printMode
-} )
-
 export const mapDispatchToProps = dispatch => ( {
   updateWindowSize: size => {
     dispatch( screenResized( size ) )
   }
 } )
 
-export default connect( mapStateToProps, mapDispatchToProps )( WindowSize )
+export default connect( null, mapDispatchToProps )( WindowSize )
 


src/WindowSize.jsx
```diff
@@ -22,11 +21,15 @@ export class WindowSize extends React.Component {
   }
 }
 
+const mapStateToProps = state => ( {
+  printMode: state.view.printMode
```
JeffreyMFarley, line 25: Is this used in the component?

src/__tests__/ActionBar.spec.jsx
```diff
@@ -15,6 +16,33 @@ describe('initial state', () => {
...
+    beforeEach(()=>{
+      jsdomOpen = window.open
+      // remember the jsdom alert
+      window.open = () => {}
+      // provide an empty implementation for window.open
+    })
+
+    afterEach(()=>{
+      window.open = jsdomOpen
+    })
```
JeffreyMFarley, line 30: 👏 👏 👏

src/actions/view.jsx
```diff
@@ -43,6 +46,30 @@ export function printModeChanged() {
...
+ *
+ * @returns {string} a packaged payload to be used by Redux reducers
+ */
+export function printModeOff() {
+  return {
+    type: PRINT_MODE_OFF,
+    requery: REQUERY_NEVER
+  }
+}
+
```
JeffreyMFarley, line 72: I prefer the separate methods rather than `toggle` because of [idempotency](https://en.wikipedia.org/wiki/Idempotence#Computer_science_meaning)

printModeOff x N = printMode: Off

printModeToggle x N = ????

JeffreyMFarley:
I also appreciate how, after closing, it goes back to normal operation in the new window

src/WindowSize.jsx
flacoman91: whoops, leftover from the loads of things i threw at this for testing

src/actions/view.jsx
flacoman91: definitely easier to test and we know what to expect. 👍

PR #238, opened 2020-04-20
Disable mouseover tooltip in print mode

flacoman91:
Disable tooltip
Disable mouseover events
disable hover style

disable-events


sephcoster:
Works on my machine. 👍

PR #240, opened 2020-04-23
chore: rename date interval to date range

flacoman91:
In the upcoming Trends Interval is defined as Day, Week, Month, Quarter, Year

We need to rename Range to Interval to be more consistent.


JeffreyMFarley:
This is a good idea to fix a name collision before proceeding

PR #249, opened 2020-05-01
Adding Trends actions 

flacoman91:
Keeping the PRs small in logical, manageable chunks

This PR adds the actions we need for the Trends Panel refinement functionality.  

Allows us to select:
- Data Lens (Overall, Company, Product, Issue)
- Date interval (Day, Week, Month, Quarter Year)

Also removed the Map related side-effects when we're not on the Map tab

### DIFF
--- a/src/reducers/__tests__/query.spec.jsx
+++ b/src/reducers/__tests__/query.spec.jsx
@@ -519,6 +519,20 @@ describe( 'reducer:query', () => {
           } )
         } );
 
+        it( 'handles when present - Map', () => {
+          action.filterName = 'has_narrative'
+          const state = {
+            has_narrative: true,
+            tab: types.MODE_MAP
+          }
+          expect( target( state, action ) ).toEqual( {
+            enablePer1000: true,
+            mapWarningEnabled: true,
+            queryString: '?tab=Map',
+            tab: types.MODE_MAP
+          } )
+        } )
+
         it( 'handles when absent' , () => {
           action.filterName = 'has_narrative'
           const state = {}


src/reducers/__tests__/query.spec.jsx
```diff
@@ -711,6 +711,38 @@ describe( 'reducer:query', () => {
...
+      it( 'changes the dataLens', () => {
+        const action = {
+          type: actions.DATA_LENS_CHANGED,
+          dataLens: 'foo'
+        }
+        const result = target( {}, action )
+        expect( result ).toEqual( {
+          dataLens: 'foo',
+          enablePer1000: true,
+          mapWarningEnabled: true,
```
JeffreyMFarley, line 724: May want to think about this.  It's not really a filter, so it shouldn't affect the per1000 warning when they toggle back to the map tab

src/reducers/__tests__/query.spec.jsx
```diff
@@ -711,6 +711,38 @@ describe( 'reducer:query', () => {
...
+      it( 'changes the dateInterval', () => {
+        const action = {
+          type: actions.DATE_INTERVAL_CHANGED,
+          dateInterval: 'Day'
+        }
+        const result = target( {}, action )
+        expect( result ).toEqual( {
+          dateInterval: 'Day',
+          enablePer1000: true,
+          mapWarningEnabled: true,
```
JeffreyMFarley, line 740: Same.   Maybe we want a Trends reducer?

flacoman91: I do have a Trends reducer in the trends work I'm doing.  

These map params comes from function that run after all of the filter actions are run.  I'll add some logic in there so this stuff does not added to the state when it's a trends related action

src/reducers/__tests__/query.spec.jsx
```diff
@@ -486,8 +515,6 @@ describe( 'reducer:query', () => {
             has_narrative: true
           }
           expect( target( state, action ) ).toEqual( {
-            enablePer1000: true,
-            mapWarningEnabled: true,
```
JeffreyMFarley, line 490: has_narrative is a filter

src/reducers/__tests__/query.spec.jsx
```diff
@@ -584,24 +627,52 @@ describe( 'reducer:query', () => {
 
       it( "adds all filters if they didn't exist", () => {
         expect( target( {}, action ) ).toEqual( {
-          enablePer1000: false,
```
JeffreyMFarley, line 587: 🤔

src/reducers/__tests__/query.spec.jsx
flacoman91: I just added a test for when it's in map mode.

src/reducers/__tests__/query.spec.jsx
flacoman91: There's a separate test for this when it's in Map mode.

JeffreyMFarley: I see how this is working now.  I went to the file itself.  The diffs weren't making sense

PR #251, opened 2020-05-12
Enforce "Last Indexed Date" as the maximum date filter

JeffreyMFarley:
This pull request is one of three PRs required to implement a new feature.

- [CCDB Data Pipeline](https://github.com/cfpb/ccdb-data-pipeline/pull/48)
- [CF GOV Refresh](https://github.com/cfpb/cfgov-refresh/pull/5723)

## Problem Statement

Occasionally, the daily complaint pipeline does not run.

When that happens, the Consumer Complaint database application will allow the user to enter a "Through" date that is later than available records.

##### Example:

June 12 - The overnight pipeline fails or is unable to complete its run.

The last indexed date would be "June 11", since that was the last successful run.

If a user launches CCDB, they will see "6/11/2020" as the last indexed date, but they can still set "6/12/2020" as the "Through" date.

## Product Owner's request

1. Do not allow the user to select a "Through" date that is greater than the "Last Indexed Date"

## Solution

Since we want to avoid multiple calls between the UI and the API to determine the last indexed date, we have chosen this method instead:

1. The nightly pipeline writes out a version of `metadata.json` that can be directly loaded as Javascript

    ```Javascript
    var complaint_public_metadata = {
      "metadata_timestamp": "2020-05-12 03:07:02",
      "qas_timestamp": "2020-05-11 19:07:02",
      "total_count": 2614976
    }
    ```

1. This script is loaded as a global variable as when the `index.html` is loaded

    ```HTML
    <script src="https://files.consumerfinance.gov/ccdb/metadata.js" type="text/javascript"></script>
    ```

1. During startup, the React application uses the value in the metadata to determine the maximum date
1. The first API call is made with respect to the metadata indexed date

## Testing

- 💯 

## Checklist


JeffreyMFarley:
> Can we change the target branch to Master to resolve in production first?

Rich did a major structure refactor, but I don't think any _features_ are in there, so it might be OK

JeffreyMFarley:
> Will the reference to the metadata file change to something permanent?

Yeah, once the pipeline PR is merged and runs in prod once

flacoman91:
> > Can we change the target branch to Master to resolve in production first?
> 
> Rich did a major structure refactor, but I don't think any _features_ are in there, so it might be OK

I'm pretty sure it's just refactored code and no new features were added.

PR #258, opened 2020-05-27
Trends Actions & tests

flacoman91:
This PR breaks out the actions into a separate PR

restructured the map reducer to place the aggregation results under results to keep things a bit simpler for trends.

Updated test coverage


src/actions/complaints.jsx
```diff
@@ -54,9 +57,9 @@ export function sendHitsQuery() {
       case MODE_MAP:
         dispatch( getStates() )
         break
-      // case 'Trends':
-      //   dispatch( getTrends() )
-      //   break
+      case MODE_TRENDS:
```
sephcoster, line 60: 👍

PR #263, opened 2020-05-28
external tooltip

flacoman91:
adding external tooltip for trends tab.

### DIFF
--- a/src/components/__tests__/__snapshots__/ExternalTooltip.spec.jsx.snap
+++ b/src/components/__tests__/__snapshots__/ExternalTooltip.spec.jsx.snap
@@ -10,8 +10,7 @@ exports[`initial state renders without crashing 1`] = `
     className="a-micro-copy"
   >
     <span>
-      Date interval: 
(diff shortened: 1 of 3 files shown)


src/components/__tests__/__snapshots__/ExternalTooltip.spec.jsx.snap
```diff
@@ -0,0 +1,68 @@
...
+exports[`initial state renders without crashing 1`] = `
+<section
+  className="tooltip-container u-clearfix"
+>
+  <p
+    className="a-micro-copy"
+  >
+    <span>
+      Date interval: 
+      ext tip title
```
JeffreyMFarley, line 14: Maybe the title label also needs to be passed in?  Or just make the title dump a single string

flacoman91: Good idea.  Sometimes that Date Interval text isn't applicable, for instance a Single Day

PR #283, opened 2020-06-16
Remove Issues from Trends & Map

flacoman91:
💥  issues

remove-issue


src/components/Trends/TrendsPanel.jsx
```diff
@@ -164,8 +156,6 @@ const mapStateToProps = state => {
   const {
     company: companyFilters,
     dateInterval,
-    issue: issueFilters,
```
flacoman91, line 167: it's funny that this thing wasn't being used for anything.  I did a doubletake once i looked at why focusData was passing in issueFilters into the function.  Probably some legacy code from explorer.  🤷

PR #287, opened 2020-06-22
View more Agg toggle under row chart

flacoman91:
adding this thing under the row chart

Screen Shot 2020-06-22 at 7 41 26 AM

- removed total from external tooltip when line chart selected

view-more

### DIFF
--- a/src/components/Trends/TrendDepthToggle.jsx
+++ b/src/components/Trends/TrendDepthToggle.jsx
@@ -16,7 +16,7 @@ export class TrendDepthToggle extends React.Component {
                     this.props.increaseDepth( diff )
                   } }>
             <span className={ 'plus' }></span>
-            View more
+            Show more
           </button>
         </div>
       }
@@ -27,7 +27,7 @@ export class TrendDepthToggle extends React.Component {
                   this.props.resetDepth()
                 } }>
           <span className={ 'minus' }></span>
-          View fewer
+          Show less
         </button>
       </div>
     }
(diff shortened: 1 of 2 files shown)


src/components/Trends/TrendDepthToggle.jsx
```diff
@@ -0,0 +1,66 @@
...
+        </div>
+      }
+      return <div className={ 'trend-depth-toggle' }>
+        <button className={ 'a-btn a-btn__link' }
+                id={ 'trend-depth-button' }
+                onClick={ () => {
+                  this.props.resetDepth()
+                } }>
+          <span className={ 'minus' }></span>
+          { `View ${ resetCount } fewer ${ lens.toLowerCase() }s` }
```
JeffreyMFarley, line 30: "View 13 fewer" feels awkward to me.  Was this in Matt's comp?  I would think "View only 5" or simply "View fewer" would be better

src/reducers/query.jsx
```diff
@@ -31,6 +31,7 @@ export const defaultQuery = {
   subLens: '',
   tab: types.MODE_MAP,
   totalPages: 0,
+  trend_depth: 5,
```
JeffreyMFarley, line 34: 🐫  case or 🐍  case?

flacoman91: should be camel case for JS standards.  the api wants the 🐍  though

src/components/Trends/TrendDepthToggle.jsx
flacoman91: We decided on Show More / Show Fewer in the call.

PR #291, opened 2020-06-25
Expand map rows

flacoman91:
Here it is:

expando-rows-map

### DIFF
--- a/src/reducers/trends.jsx
+++ b/src/reducers/trends.jsx
@@ -6,8 +6,8 @@ import * as colors from '../constants/colors'
 import {
   clamp, coalesce, getSubKeyName, processErrorMessage
 } from '../utils'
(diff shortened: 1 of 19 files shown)


src/reducers/trends.jsx
```diff
@@ -304,6 +269,16 @@ export const getColorScheme = ( lens, rowNames ) => {
...
+/**
+ * helper function validate payload and also get under eslint limit
+ * @param {object} aggregations payload from api
+ * @param {string} k the key we need to validate against
+ * @returns {boolean} whether the bucket is valid
+ */
+export function validateBucket( aggregations, k ) {
+  return aggregations[k] && aggregations[k].doc_count &&
+    aggregations[k][k] && aggregations[k][k].buckets
+}
```
JeffreyMFarley, line 281: :+1:

PR #302, opened 2020-07-08
prevent britecharts mutate

flacoman91:


flacoman91:
never mind, this thing doesnt work

PR #303, opened 2020-07-08
fix date range empty values

flacoman91:
- fixing bug where Row Chart was mutating redux (using a deep clone)
- Backfilling zeroes when there's no values using new API dateRangeBuckets object
- cleaned up some of the reducer, removing legacy code with pctOfSet since it's not needed anymore
- simplified the processTrends function in the reducer so it's more explicit as to what gets processed

needs API update

Screen Shot 2020-07-08 at 5 02 31 PM
Screen Shot 2020-07-08 at 5 02 20 PM
Screen Shot 2020-07-08 at 5 01 47 PM
Screen Shot 2020-07-08 at 5 01 36 PM
Screen Shot 2020-07-08 at 5 01 31 PM


src/components/Charts/RowChart.jsx
```diff
@@ -89,8 +89,9 @@ export class RowChart extends React.Component {
     const {
       colorScheme, data, id, printMode, toggleRow, total
     } = this.props
-
-    const rows = data.filter( o => {
+    // deep copy
+    // do this to prevent REDUX pollution
+    const rows = JSON.parse( JSON.stringify( data ) ).filter( o => {
```
JeffreyMFarley, line 94: This is pretty clever!  What does the internet say is the best way to deep clone in Javascript?

Even if it is a little expensive, `_redrawChart` is only called when the values change and I don't think `data` is that deep for a row chart

src/reducers/__fixtures__/trendsAggs.jsx
```diff
@@ -1 +1 @@
...
..."value": -21 } }, { "key_as_string": "2020-04-01T00:00:00.000Z", "key": 1585699200000, "doc_count": 22, "interval_diff": { "value": 15 } }, { "key_as_string": "2020-03-01T00:00:00.000Z", "key": 1583020800000, "doc_count": 7 } ] } }, { "key": "Older American, Servicemember", "doc_count": 7, "trend_period": { "buckets": [ { "key_as_string": "2020-05-01T00:00:00.000Z", "key": 1588291200000, "doc_count": 1, "interval_diff": { "value": -4 } }, { "key_as_string": "2020-04-01T00:00:00.000Z", "key": 1585699200000, "doc_count": 5, "interval_diff": { "value": 4 } }, { "key_as_string": "2020-03-01T00:00:00.000Z", "key": 1583020800000, "doc_count": 1 } ] } } ] } }, "dateRangeBuckets": { "doc_count": 121054, "dateRangeBuckets": { "buckets": [ { "key_as_string": "2020-01-01T00:00:00.000Z", "key": 1577836800000, "doc_count": 21519 }, { "key_as_string": "2020-02-01T00:00:00.000Z", "key": 1580515200000, "doc_count": 25096 }, { "key_as_string": "2020-03-01T00:00:00.000Z", "key": 1583020800000, "doc_count": 29506 }, { "key_as_string": "2020-04-01T00:00:00.000Z", "key": 1585699200000, "doc_count": 35112 }, { "key_as_string": "2020-05-01T00:00:00.000Z", "key": 1588291200000, "doc_count": 9821 } ] } } }
```
JeffreyMFarley, line 1: Is this parseable as a JSON block?   I had problems trying to run an JSON prettifier on it when I was trying to change the values

src/reducers/__tests__/trends.spec.jsx
```diff
@@ -229,28 +230,24 @@ describe( 'reducer:trends', () => {
...
       result = target( state, action )
       expect( result ).toEqual( trendsAggsDupeResults )
     } )
 
     it( 'maps data to object state - Missing Bucket', () => {
+      // to replicate this
+      // ?date_received_max=2017-07-08
+      // &date_received_min=2017-03-08
+      // &from=0&lens=Product&tab=Trends
+      // you'll get broken buckets since the product recategorization in apr
```
JeffreyMFarley, line 250: 🥇

src/reducers/trends.jsx
```diff
@@ -118,31 +118,37 @@ export function mainNameLens( lens ) {
  * processes the stuff for the area chart, combining them if necessary
  * @param {object} state redux state
  * @param {object} aggregations coming from the trends api
- * @param {Array} buckets contains trend_period data
  * @returns {object} the data areas for the stacked area chart
  */
-function processAreaData( state, aggregations, buckets ) {
+function processAreaData( state, aggregations ) {
```
JeffreyMFarley, line 123: Thanks for fixing this

src/reducers/trends.jsx
```diff
@@ -315,43 +323,26 @@ export function validateBucket( aggregations, k ) {
...
+
+  results.dateRangeLine = processLineData( lens, aggregations, focus, subLens )
+
+  // add issue or whatever key you need in if there's more
+  const keys = [ 'company', 'product' ]
+  keys.forEach( k => {
+    results[k] = aggregations[k] ?
+      processBucket( state, aggregations[k][k].buckets ) : []
+  } )
+
```
JeffreyMFarley, line 345: Yay for reducing complexity!

src/components/Charts/RowChart.jsx
flacoman91: that's the way. I tried the es6 spread operator, array.from, but it would still get mutated since it was a new array, but the objects inside were still pointers to the original object and get modified by britecharts.

src/reducers/__fixtures__/trendsAggs.jsx
flacoman91: yes. 

prettify

PR #312, opened 2020-07-13
fix expando chart bug

flacoman91:
Fixing bug where pasting in URL and then switching between tabs weren't allowing certain rows to expand or collapse properly. 

When we start from fresh (not from a pasted url), anytime we do any changes both reducers are in sync. 

When we paste in a url, the map and trends reducers aren't in sync, so it looks like if you close on one tab/reducer, it's forcing it open on the other reducer.  Clearing out the results and available filters when the tab changes prevents this bug from happening.

Also hiding the More Information about... text on the row chart when printing.

### DIFF
--- a/src/reducers/trends.jsx
+++ b/src/reducers/trends.jsx
@@ -424,7 +424,8 @@ function updateRowVisibility( state, expandedTrends ) {
     if ( results.hasOwnProperty( k ) && Array.isArray( results[k] ) ) {
       results[k]
         .forEach( o => {
-          o.visible = Boolean( expandedTrends.includes( o.parent ) || o.isParent )
+          o.visible =
+            Boolean( expandedTrends.includes( o.parent ) || o.isParent )
         } )
     }
   }


src/reducers/map.jsx
```diff
@@ -46,6 +46,21 @@ export const processStateAggregations = agg => {
...
+ * Updates the state when an tab changed occurs, reset values to start clean
+ *
+ * @param {object} state the current state in the Redux store
+ * @returns {object} the new state for the Redux store
+ */
+export function handleTabChanged( state ) {
+  return {
+    ...state,
+    expandedTrends: [],
+    filterNames: [],
```
JeffreyMFarley, line 59: I missed something. What are filters being reset  when the tabs vhange?

flacoman91: these are just the parent rows from the row chart that are clickable

JeffreyMFarley: That seems like a more appropriate name

flacoman91: updated for clarity.  It was leftover code from Explore

flacoman91:
sorry, I want to update some things to make them more idempotent

flacoman91:
Interesting as to why this bug cropped up.

Since you have these values in both map and trend reducer, when you paste in the url, only the map reducer gets the value with rows already open.  

You then switch to the trends tab, and click something to open it, the old toggle action flips it to open, but modifies the map reducer as well to close, and it messes up the data when you switch back to trends and you can't do anything.  

The idempotent actions gets rid of this and tell us, just expand or collapse.  It's way simpler and error prone.

PR #319, opened 2020-07-16
show more/less toggle to appear on overview

flacoman91:
fixing toggle to appear on overview page

### DIFF
--- a/src/reducers/query.jsx
+++ b/src/reducers/query.jsx
@@ -789,7 +789,8 @@ function changeFocus( state, action ) {
     focus,
     lens,
     subLens: state.subLens || getSubLens( lens ),
-    tab: types.MODE_TRENDS
+    tab: types.MODE_TRENDS,
+    trendDepth: 25
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/reducers/query.jsx
```diff
@@ -998,6 +1009,8 @@ export default ( state = defaultQuery, action ) => {
     validateDateInterval( newState )
   }
 
+  // remove any filter keys with empty array
+  pruneEmptyFilters( newState )
```
JeffreyMFarley, line 1013: Wait... what happens in this sequence

1. Filter on `90210`
1. Filter on State = `Kansas`

Both filters get dropped?

flacoman91: doesn't seem like its getting dropped from the query reducer
filter

JeffreyMFarley:
```
    zip_code: [
      '90210'
    ],
    state: [
      'KS'
    ]
```

So the filter arrays have values, but the counts are zero.. gotcha

sephcoster: Thanks for the clarification!

PR #321, opened 2020-07-16
- Show NN less / +Show NN more text fix

flacoman91:
Fixing +/- sign under text for the MoreOrLess toggle

plus-minus

### DIFF
--- a/src/components/Filters/__tests__/MoreOrLess.spec.jsx
+++ b/src/components/Filters/__tests__/MoreOrLess.spec.jsx
@@ -1,7 +1,12 @@
-import React from 'react'
-import { shallow } from 'enzyme'
 import AggregationItem from '../AggregationItem'
+import configureMockStore from 'redux-mock-store'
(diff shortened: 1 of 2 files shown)


src/components/Filters/MoreOrLess.jsx
```diff
@@ -45,7 +45,9 @@ export class MoreOrLess extends React.Component {
           <div>
                <button className="a-btn a-btn__link more"
                        onClick={ this._toggleShowMore }>
-                  + Show {remain} {this.state.showMore ? 'less' : 'more'}
+                 { this.state.showMore ? `- Show ${ remain } less` :
```
anselmbradford, line 48: I _think_ this would work too?

```js
{
  `- Show ${ remain } ${ this.state.showMore ? 'less' : 'more'  }`
}
```

JeffreyMFarley: We also want a `+` when `More`

PR #323, opened 2020-07-20
validate dateInterval and dateRange url parameters

flacoman91:
Fixing bug where bad params was breaking the Trends

incorrect-param

### DIFF
--- a/src/reducers/__tests__/query.spec.jsx
+++ b/src/reducers/__tests__/query.spec.jsx
@@ -298,6 +298,13 @@ describe( 'reducer:query', () => {
       expect( actual.size ).toEqual( 100 )
     } )
 
+    it( 'handles bogus date parameters', () => {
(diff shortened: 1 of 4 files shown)


src/reducers/__tests__/query.spec.jsx
```diff
@@ -298,6 +298,13 @@ describe( 'reducer:query', () => {
       expect( actual.size ).toEqual( 100 )
     } )
 
+    it( 'handles bogus date parameters', () => {
+      action.params = { dateInterval: '3y', dateRange: 'Week' }
+      const actual = target( state, action )
+      expect( actual.dateInterval ).toEqual( 'Month' )
+      expect( actual.dateRange ).toEqual( '3y' )
+    } )
+
```
JeffreyMFarley, line 307: 🏅

PR #326, opened 2020-07-21
Page spacing improvements, misc bug fixes

flacoman91:
General bug fixes and improvements: 
- remove line on right side of content area
- adjust left side of charts to left align things
- adjust drop-down selects on phone to make them full width
- Fixing legend incorrect colors. now matches the row, area and line charts
Screen Shot 2020-07-21 at 3 32 51 PM
Screen Shot 2020-07-21 at 3 32 44 PM

- Print view - do not render "visualize trends" text from row chart instead of hiding with CSS to improve spacing
Screen Shot 2020-07-21 at 3 33 32 PM

- ChartType and Lens URL / Option validation in query and trends reducer - chartType defaults to line if 'Overview' is selected
url-check

- Adding padding to text, filters, select boxes, gutters depending on breakpoints
  - Ultrawide
Screen Shot 2020-07-21 at 3 20 20 PM
Screen Shot 2020-07-21 at 3 20 11 PM

  - Superwide
Screen Shot 2020-07-21 at 3 19 54 PM

  - Desktop

Screen Shot 2020-07-21 at 3 21 16 PM
Screen Shot 2020-07-21 at 3 20 53 PM
Screen Shot 2020-07-21 at 3 20 43 PM
Screen Shot 2020-07-21 at 3 20 37 PM

  - Tablet
Screen Shot 2020-07-21 at 3 21 50 PM
Screen Shot 2020-07-21 at 3 21 43 PM
Screen Shot 2020-07-21 at 3 21 35 PM

  - Phone
Screen Shot 2020-07-21 at 3 22 24 PM
Screen Shot 2020-07-21 at 3 22 10 PM
Screen Shot 2020-07-21 at 3 22 03 PM

### DIFF
--- a/src/components/Charts/LineChart.less
+++ b/src/components/Charts/LineChart.less
@@ -17,17 +17,6 @@
     .masking-rectangle {
         display: none;
     }
-
-    /*primary comparable company */
-
-    .chart-group {
-        .topic:nth-child(2),
-        .topic:nth-child(3) {
-            .line {
-                stroke-width: 2;
-            }
-        }
-    }
 }
 
 #stacked-area-chart,


src/components/Charts/RowChart.jsx
```diff
@@ -93,7 +93,8 @@ export class RowChart extends React.Component {
     // deep copy
     // do this to prevent REDUX pollution
     const rows = cloneDeep( data ).filter( o => {
-      if ( o.name ) {
+      if ( o.name && printMode ) {
```
JeffreyMFarley, line 96: > Print view - do not render "visualize trends" text from row chart instead of hiding with CSS to improve spacing

I would check the truth table on this.  It is only true when o.name AND printMode

flacoman91: that's intentional.   I have some unit tests that don't have "name" set in the input objects (pretty much just an array of anything)

Normally we want to just render / pass everything to the row chart, and here I want to filter the data.

flacoman91:
Updated screens after matt's feedback

Screen Shot 2020-07-22 at 12 46 03 PM
Screen Shot 2020-07-22 at 12 45 56 PM
Screen Shot 2020-07-22 at 12 45 18 PM
Screen Shot 2020-07-22 at 12 43 45 PM

sephcoster:
It looks like the line from the overview chart is retaining its stroke width on the Product / Aggregation line charts. Could we restore it to its normal width for those pages so it doesn't over-weight a single line out of five?

PR #328, opened 2020-07-24
Polish up code base

flacoman91:
- move expandedRows to view reducer
- simplify logic
- updating defaults when aggs received
- clearing out results
  - When users select filters to give us zero results
  - When filters change
  - when tab changes
- explicit definitions of which aggs we need based on Lens/Focus
- reset date range to 3y when filter all removed

### DIFF
--- a/src/reducers/__tests__/query.spec.jsx
+++ b/src/reducers/__tests__/query.spec.jsx
@@ -627,26 +627,22 @@ describe( 'reducer:query', () => {
       it( 'clears all filters', () => {
         const actual = target( state, action )
         expect( actual ).toMatchObject( {
-          dateRange: 'All',
+          dateRange: '3y',
           from: 100,
           searchField: 'all',
           size: 100
         } )
 
-        expect( actual.queryString ).toContain( 'dateRange=All' )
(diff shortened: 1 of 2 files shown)


src/reducers/trends.jsx
```diff
@@ -4,38 +4,61 @@
...
+  {
+    chartType: 'line',
+    focus: '',
+    lens: 'Overview',
+    subLens: ''
+  },
+  { ...getResetState() }
+)
+
+export const defaultState = getDefaultState()
```
JeffreyMFarley, line 42: nice

src/reducers/view.jsx
```diff
@@ -18,8 +20,11 @@ export const defaultView = {
 function processParams( state, action ) {
   const params = action.params
 
-  state.printMode = params.printMode === 'true' ? true : false
-  state.fromExternal = params.fromExternal === 'true' ? true : false
+  state.printMode = params.printMode === 'true'
+  state.fromExternal = params.fromExternal === 'true'
```
JeffreyMFarley, line 24: :D

PR #329, opened 2020-07-28
more polish

flacoman91:
- fixing undefined props in snapshot tests
- sanitize html ids in dynamic properties to follow [html spec](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id#:~:text=id%20's%20value%20must%20not,have%20one%20single%20ID%20value.)

### DIFF
--- a/src/components/Filters/AggregationBranch.jsx
+++ b/src/components/Filters/AggregationBranch.jsx
@@ -91,7 +91,7 @@ export class AggregationBranch extends React.Component {
                  disabled={item.disabled}
                  checked={checkedState === CHECKED}
                  className="flex-fixed a-checkbox"
-                 id={ sanitizeHtmlId( id ) }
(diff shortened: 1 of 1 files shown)


src/components/Filters/AggregationBranch.jsx
```diff
@@ -85,7 +91,7 @@ export class AggregationBranch extends React.Component {
                  disabled={item.disabled}
                  checked={checkedState === CHECKED}
                  className="flex-fixed a-checkbox"
-                 id={id}
+                 id={ sanitizeHtmlId( id ) }
```
JeffreyMFarley, line 94: You sanitize it twice?  It better be clean now!

flacoman91: ah whoops. that was an accident

PR #333, opened 2020-07-29
cleaning up url based on tab, updated per1000 complaints logic

flacoman91:
limited valid parameters on:
- front end
- API calls

Updated logic for enable per 1000 complaints:
- allow per 1000 when no filters, or only state filter selected
- disable filters per usual when other filters selected
per1000

urltidy


flacoman91:
putting this on hold since i noticed we can do the same thing for the API component of this.

PR #338, opened 2020-08-05
prevent action from firing when same

flacoman91:
fixes the following actions from firing when the value is the same:
- dateRange
- chartType

Update tests using Redux[Component] syntax with mockStore
- chartToggles
- dateRanges

### DIFF
--- a/src/components/__tests__/ChartToggles.spec.jsx
+++ b/src/components/__tests__/ChartToggles.spec.jsx
@@ -1,5 +1,5 @@
 import configureMockStore from 'redux-mock-store'
-import {
+import ReduxChartToggles, {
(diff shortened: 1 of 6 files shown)


src/components/__tests__/ChartToggles.spec.jsx
```diff
@@ -1,22 +1,30 @@
 import configureMockStore from 'redux-mock-store'
-import {
+import ReduxChartToggles, {
```
JeffreyMFarley, line 2: Glad to see you are making progress on fixing the snapshot tests

flacoman91: thanks. eventually we'll get it done..

PR #342, opened 2020-08-12
Updated Filter logic

flacoman91:
Updated logic for filters with Parent/Child relationships (issue/product)

Check only parent when all available children are checked
Uncheck parent and apply the rest of the siblings when a parent is checked, and a child item is "unchecked"

Fake "Checked" subitems appear on the front-end, but do not apply it in the query or in redux

Don't apply fake pills

### DIFF
--- a/src/utils/__tests__/compare.spec.jsx
+++ b/src/utils/__tests__/compare.spec.jsx
@@ -13,6 +13,11 @@ describe( 'arrayEquals', () => {
     expect( res ).toBeTruthy()
   } )
 
+  it( 'compares unsorted arrays', () => {
+    const res = sut.arrayEquals( [ 2, 1 ], [ 1, 2 ] )
+    expect( res ).toBeFalsy()
+  } )
+
   it( 'compares arrays with different values', () => {
     const res = sut.arrayEquals( [ 2, 2 ], [ 1, 2 ] )
     expect( res ).toBeFalsy()


src/components/Filters/AggregationItem.jsx
```diff
@@ -40,24 +51,82 @@ AggregationItem.propTypes = {
...
 
 export const mapStateToProps = ( state, ownProps ) => {
-  const field = state.query[ownProps.fieldName]
+  const aggs = coalesce( state.aggs, ownProps.fieldName, [] )
+  const filters = coalesce( state.query, ownProps.fieldName, [] )
   const value = ownProps.item.key
-  const active = typeof field !== 'undefined' && field.indexOf( value ) > -1
-  return { active }
+  const parentKey = value.split( SLUG_SEPARATOR )[0]
+  const active = filters.includes( value ) || filters.includes( parentKey )
```
JeffreyMFarley, line 66: clever

src/components/Filters/__tests__/AggregationBranch.spec.jsx
```diff
@@ -127,21 +127,32 @@ describe('component::AggregationBranch', () => {
       const checkbox = target.find('li.parent input[type="checkbox"]')
       checkbox.simulate('change')
       expect(props.uncheckParent).not.toHaveBeenCalled()
-      expect(props.checkParent).toHaveBeenCalledWith(
-        'issue', ['foo', 'foo•bar', 'foo•baz', 'foo•qaz', 'foo•quux']
-      )
+      expect(props.checkParent).toHaveBeenCalled()
```
JeffreyMFarley, line 130: Are you not able to test with expected values anymore?

flacoman91: the function is passing in props, so it's also adding in some js functions so it looks a bit cluttered in this test.  i can explicitly check for these

src/utils/__tests__/compare.spec.jsx
```diff
@@ -0,0 +1,20 @@
...
+// ----------------------------------------------------------------------------
+// Tests
+describe( 'arrayEquals', () => {
+  it( 'compares different length arrays', () => {
+    const res = sut.arrayEquals( [], [ 1, 2 ] )
+    expect( res ).toBeFalsy()
+  } )
+
+  it( 'compares equal arrays', () => {
+    const res = sut.arrayEquals( [ 1, 2 ], [ 1, 2 ] )
```
JeffreyMFarley, line 12: Does element order matter?

flacoman91: yes. I just added another test to verify a different order.

sephcoster:
Great work, thanks!

PR #356, opened 2021-06-30
1478: es7 update

flacoman91:
- fixes the pager counts
- reset pager & requery when the following changes: 
  - sort
  - size
  - Search Text
  - Search Field
- Added Cypress testing for List view, fixed broken cypress tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1186,95 +1186,6 @@
       "resolved": "https://registry.npmjs.org/@csstools/normalize.css/-/normalize.css-9.0.1.tgz",
       "integrity": "sha512-6It2EVfGskxZCQhuykrfnALg7oVeiI6KclWSmGDqB0AiInVrTGB9Jp9i4/Ad21u9Jde/voVQz6eFX/eSg/UsPA=="
     },
-    "@cypress/listr-verbose-renderer": {
-      "version": "0.4.1",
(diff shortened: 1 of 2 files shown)


src/actions/complaints.jsx
```diff
@@ -105,7 +105,7 @@ export function getAggregations() {
 export function getComplaints() {
   return ( dispatch, getState ) => {
     const store = getState()
-    const qs = store.query.queryString + '&no_aggs=true'
```
higs4281, line 108: Do we need aggregations in the complaint request, or was this just to get the request cycle working? 
Not running aggregations on every request is a bit more efficient. Not huge.

flacoman91:
I removed it to get it running.

higs4281:
Thanks. 

We also need aggregations to find the total search count when there are 10K or more hits, so it makes sense to have them included.

setup.py
```diff
@@ -39,7 +39,7 @@ def read_file(filename):
 
 setup(
     name='ccdb5_ui',
-    version=get_git_version(),
+    version='2.3.3-dev5',
```
higs4281, line 42: Not related to your PR, but could we revert? I made this change while struggling to get ccdb deployed to DEV5, and we can switch back now.

```suggestion
    version=get_git_version(),
```

higs4281 (approved):
It's looking good from here.

PR #357, opened 2021-07-08
ES7 update

contolini:
Continuation of https://github.com/cfpb/ccdb5-ui/pull/356

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ def read_file(filename):
 
 setup(
     name='ccdb5_ui',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -39,7 +39,7 @@ def read_file(filename):
 
 setup(
     name='ccdb5_ui',
-    version=get_git_version(),
+    version='2.3.3-dev5',
```
higs4281, line 42: This should be reverted (my bad).

```suggestion
    version=get_git_version(),
```

PR #358, opened 2021-07-09
Isolate front-end test dependencies

higs4281:
When we deploy cfgov to production, we either get a preconfigured wheel package
or build one. Since we don't need our test rigs in production, we can
exclude them from the production build by moving them to a "devDependencies"
section in package.json, as the cfgov repo does.

This is a test to see if our CI runs with dependencies split up this way.


flacoman91 (approved):
Makes sense. let's try it.

PR #360, opened 2021-07-14
Unify node versioning

higs4281:
We've had trouble getting ccdb5-ui deployed to DEV servers because of
changes to package-lock.json, and we found that github-actions was using
node 10 to build wheels, and our local .nvmrc file specified 12.

Since cfgov is using 14, this change unifies node on that version and updates
the package-lock.json file to a 14 build.


higs4281:
the change to frontendbuild.sh is just to permissions to make it runnable.

contolini (approved):

PR #361, opened 2021-07-21
1478 - hide counts for Company and Zip Code

flacoman91:
With the ES7 update, we only get top 10 back in the aggregations leading to zero counts in the filters.

This PR hides the complaint counts in the zip code and company filter options.

There may be a performance hit if we try to returned doc_counts since there's over 25k zip codes and thousands of companies.

Screen Shot 2021-07-21 at 12 00 15 PM


lorenseanstewart (approved):
LGTM!

PR #363, opened 2021-07-28
Document Cypress tests and get them to run on DEV servers

higs4281:
Our Cypress integration tests cover basic user operations for the complaint search tool,  
and they are intended to be run in the context of a consumerfinance.gov website  
with an Elasticsearch index  of complaints available.

They generally run consistently locally, but running the tests against external servers proved
to be a challenge, because our API ([ccdb5-api](https://github.com/cfpb/ccdb5-api)) uses throttling
to protect against abusive requests, and the tests easily overrun the throttling limits and fail.
To avoid having to slow our tests down with a bunch of wait commands, the ccdb5-api code
Is being adjusted to turn off throttling when the host Django app has DEBUG set to True. This is   
the standard setting for local operations, and the DEV5 server is currently set up that way too.

Instructions for running local and remote Cypress tests have been added to
this repo's README.txt.

This addresses internal issue 1535.

## Testing

- On a server:  
This is most easily tested by following the README instructions for running Cypress against our DEV5 server, which uses  
Elasticsearch 7 code and has the API throttling changes applied.  Headless tests run in about 2 minutes there.
- Locally:  
You can run the tests in your local environment if you pull in the `ccdb5-cypress` branch of consumerfinance.gov and install requirements with `pip install -r requirements/libraries.txt`

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ yarn run cypress run --browser chrome --headless
 This will run Cypress against a local version of consumerfinance.gov running on port 8000. To use a different port, such as the port 3000 used by `yarn run start`, pass a `--baseUrl` parameter:
 
 ```bash
-yarn run cypress run --browser chrome --headless --baseUrl http://localhost:3000/data-research/consumer-complaints/search/
+yarn run cypress run --browser chrome --headless --config baseUrl=http://localhost:3000/data-research/consumer-complaints/search/
 ```
 
 To open a local Cypress test-runner to choose which tests to run and see the browser interactions:


cypress.json
```diff
@@ -1 +1,10 @@
-{}
+{
+    "viewportWidth": 1200,
+    "viewportHeight": 1200,
+    "responseTimeout": 80000,
+    "defaultCommandTimeout": 30000,
+    "requestTimeout": 20000,
+    "screenshotOnRunFailure": false,
+    "video": false,
+    "baseUrl": "http://localhost:8000/data-research/consumer-complaints/search/"
```
flacoman91, line 9: this is applicable if you run it through CFGOV and port 8000.  sometimes, I run ccdb-ui using yarn run start and it puts it at http://localhost:3000, so the base url would be: "baseUrl": "http://localhost:3000/"

The api still runs on port 8000.

README.md
```diff
@@ -100,10 +94,46 @@ and then run:
 tox
 ```
 
-## Known issues
+## Cypress integration tests
+
+Our broswer-based tests check base-line user operations for [consumer complaint search](https://www.consumerfinance.gov/data-research/consumer-complaints/search/). The tests are meant to be run
```
flacoman91, line 99: typo `broswer`

.gitignore
```diff
@@ -66,6 +66,9 @@ nosetests.xml
 coverage.xml
 coverage/
 
+# Cypress testing
+cypress/videos/
+
```
flacoman91, line 71: i'd also include the other cypress folders of the artifacts that could get generated during testing.

`
cypress/downloads/
cypress/screenshots/
`

lorenseanstewart (approved):
looks good pending Rich's comments

cypress.json
higs4281: Good point. We could add a note that if you use a different port for your local server, you can override settings by providing a `--baseUrl` option to the run command.

README.md
higs4281: √

.gitignore
higs4281: good catch.

cypress.json
flacoman91: Oh i see you already had a note in the readme.

flacoman91 (approved):
🥇

PR #365, opened 2021-08-06
1559 - Remove total number of pages from pager, remove submit form

flacoman91:
because of ES7 performance reasons, we remove from pager:
- text input
- form submit

Screenshots:
Screen Shot 2021-08-09 at 1 19 45 PM
Screen Shot 2021-08-09 at 1 19 06 PM
Screen Shot 2021-08-09 at 1 18 47 PM


higs4281 (approved):
This works as requested. I tried going forward and backward several pages, and results were consistent on each visit.

Not a concern for this PR, but I see no justification for paginating 10K hits now.

PR #367, opened 2021-08-25
Fixing tests and timezone for CI

flacoman91:
This PR fixes an issue where Unit tests locally and on CI don't match correctly when using the moment.js library.

We standardardize the functions to use EST here so we get consistent results.


flacoman91:
So this is what I found when i did console.log for the  timezone guess

Screen Shot 2021-08-25 at 11 18 20 AM

[Forcing the America/New York timezone](https://github.com/cfpb/ccdb5-ui/pull/367/files#diff-9c7b4e38c333c05c15fea8abd552712481f0f1aadb468016c5f0cc6c8c63a5fbR66) fixes the bug in CI and ensures the  tests are consistent locally and on the server.

lorenseanstewart (approved):

PR #368, opened 2021-09-13
1616 - fixing query filter bug

flacoman91:
resets the query state/breakpoints/page when filters or sorts change
update cypress testing


src/reducers/query.jsx
```diff
@@ -1088,6 +1100,27 @@ function handleSpecificAction( state, action ) {
...
+    actions.FILTER_REMOVED,
+    actions.FILTER_REPLACED,
+    actions.SEARCH_FIELD_CHANGED,
+    actions.SEARCH_TEXT_CHANGED,
+    actions.SIZE_CHANGED,
+    actions.SORT_CHANGED,
+    actions.TAB_CHANGED
+  ];
+  // these actions cause the page to reset to 1 and also clear out breakpts
+  if ( breakPointActions.includes( action.type ) ) {
```
flacoman91, line 1120: this is where the magic happens.  everything else is just test fixture updates

lorenseanstewart (approved):

PR #377, opened 2021-12-22
Remove Django cruft | move dependencies to devDependencies

anselmbradford:
## Changes

- Move `package.json` `dependencies` to `devDependencies` so they don't end up in the npm module.
- Remove `prepack` script, as build can target `dist` directly.
- Remove django/python infrastructure, which is no longer needed.

## Testing

- `./frontend.sh && yarn start` should launch the app.


flacoman91 (approved):
👍

PR #376, opened 2021-12-15
Set up for standalone development | release 2.4.5

anselmbradford:
## Changes

- Update github actions node version to 16.
- Add functional test step to github actions.
- Add `ccdbApiUrl` variable to the config environment.
- Pin highcharts version to prevent regression.
- Adds release scripts and steps and bumps to 2.4.5

## Testing

- `yarn run build && yarn run start` should launch the app on localhost.

PR #378, opened 2021-12-29
Add compiled dist files to repo

anselmbradford:
https://github.com/cfpb/ccdb5-ui/pull/376/ added a /dist/ directory for the compiled src code to live in, but it was gitignored. This PR undos the ignoring so that the compiled file that's distributed via npm ends up in the repo as well. This should allow Checkmarx to scan the file for issues.

## Changes

- Not gitignoring the /dist/ directory.

## Testing

- All should be as before. The /dist/ directory shows the products of `yarn run build`.


flacoman91 (approved):
👍

PR #381, opened 2022-01-05
Replace moment.js with day.js

anselmbradford:
Moment.js has been deprecated. This PR replaces it with dayjs. 

## Changes

- Replace moment.js with day.js

## Testing

- App should be unchanged and PR checks should pass.


anselmbradford:
Closing for now, this is more involved as the isValid() behaves differently in day.js https://stackoverflow.com/questions/58216504/dayjs-isvalid-behaves-differently-to-moment

PR #383, opened 2022-02-07
Prevent default behavior of button so search form doesn't submit

anselmbradford:
The search form was submitted when clicking the advanced search options toggle. This prevents that default behavior. 

## Removals

- Remove unused `o-expandable_cue` CSS classes.

## Changes

- Prevent the default event behavior when clicking the advanced search options button toggle.

## Testing

- When toggling the "Show advanced search tips" options, a search should not be performed, but the search tips should still show and hide.

### DIFF
--- a/src/components/Search/SearchBar.jsx
+++ b/src/components/Search/SearchBar.jsx
@@ -93,8 +93,10 @@ export class SearchBar extends React.Component {
               </a>
 
               <div className="advanced-container flex-fixed">
-                <button className="a-btn a-btn__link" onClick={ this._onAdvancedClicked }>
-                  { this.state.advancedShown ? 'Hide' : 'Show' } advanced search tips
+                <button className="a-btn a-btn__link"
+                        onClick={ this._onAdvancedClicked }>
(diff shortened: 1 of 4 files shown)


src/components/Search/SearchBar.jsx
```diff
@@ -95,11 +95,11 @@ export class SearchBar extends React.Component {
               <div className="advanced-container flex-fixed">
               {
                this.state.advancedShown ?
-                 <button className="a-btn a-btn__link o-expandable_cue-close"
+                  <button className="a-btn a-btn__link"
                       onClick={ this._onAdvancedClicked }>
                       Hide advanced search tips
                   </button> :
-                  <button className="a-btn a-btn__link o-expandable_cue-open"
+                  <button className="a-btn a-btn__link"
```
flacoman91, line 102: looks like the only reason why this button was set in the ternary was because of the o-expandable_cue-open classes.  maybe combine this into a single button and only have it modify the Show/Hide inner text?

anselmbradford: Good call! Updated!

flacoman91 (approved):
👍

PR #385, opened 2022-02-15
Remove incomplete periods from Line and Stacked Area charts

flacoman91:
This PR removes the last incomplete interval (Month, Week, Year, Quarter) from the line and stacked area charts.

## Additions

- updated test coverage
- added some logic to cloneDeep the values that we are passing into the line and stacked area charts.  Britecharts has a tendancy to mutate the values we have in redux (big no-no!).   In the Explorer react refactor work, we use redux toolkit, which has [immer](https://immerjs.github.io/immer/) included and it enforces the immutable objects and this bug was caught. 
- added warning when user selects date ranges that will now allow chart to render. (trim the incomplete start and end date, i.e. pick 1 yr interval but date filter is under 1 year)

## Removals

- removed lastDate from the trends reducer.  We previously set this when the trends_received from the api, but with this new logic, we have to manually calculate the last date depending on if we need to remove incomplete periods.  We used the lastDate value to figure out which date to dispatch an updated external tooltip for the line and stacked area charts.

## Changes

-

## Testing

- to test this, just build normally and play with the "To" date filter.  For instance, if you set the interval to "Year" and the "To" date as 12/31/2021, you should see the chart display that final period.    If you set the "To" date to 12/15/2021, you won't see any incomplete periods for the year 2021 and the last period should be ending on 12/31/2020.

## Screenshots
Screen Shot 2022-02-17 at 9 30 23 PM
Screen Shot 2022-02-17 at 9 30 36 PM
Screen Shot 2022-02-17 at 9 30 30 PM

## Notes

-

## Todos

-

## Checklist


src/components/Charts/LineChart.jsx
```diff
@@ -7,9 +7,11 @@ import { line, tooltip } from 'britecharts'
 import { connect } from 'react-redux'
 import { hashObject } from '../../utils'
 import { isDateEqual } from '../../utils/formatDate'
+import moment from 'moment';
```
anselmbradford, line 10: It makes sense to use moment since it's used in the app already, but since moment is deprecated we should transition at some point to dayjs. If you see anything that makes that transition easier while working with moment that'll probably help down the road.

flacoman91:
> Checking on expected behavior here… so with a range of 2/1/2022-2/16/2022 I get an empty chart area:
> 
> Screen Shot 2022-02-17 at 9 53 02 AM
> 
> With 1/1/2022-2/16/2022, I get an empty chart: Screen Shot 2022-02-17 at 9 53 41 AM
> 
> Shouldn't these two be the same? (Either show an empty chart or show nothing?)

you're right. it should show nothing. 

In your first example, both dates were chopped off, so there was no data to plot.

In the 2nd example, only 1 date was valid, so it was trying to plot a single dot, but for a line, you need 2 points (dates) to plot against.

anselmbradford:
Would it be easy to hide the "Complaint" and "Data received by the CFPB" text when the chart doesn't show up?

Screen Shot 2022-02-17 at 9 53 02 AM

flacoman91:
> Would it be easy to hide the "Complaints" and "Data received by the CFPB" text when the chart doesn't show up?
> 
> Screen Shot 2022-02-17 at 9 53 02 AM

You'd have to do good amount of restructuring of the code to get that to work.  If it were a functional component like we have in explorer-react, it would be trivial, since the render happens after all of the calculations are done.

The way this code here works is that the redrawChart does the cleanup if the data is incorrect and the render function has the parent container which has that complaints text.  MapStatetoProps is a bit backwards in the way it's done.

anselmbradford:
> The way this code here works is that the redrawChart does the cleanup if the data is incorrect and the render function has the parent container which has that complaints text. MapStatetoProps is a bit backwards in the way it's done.

What if we added a `u-hidden` class to the chart container when it was to be hidden?

anselmbradford (approved):

PR #386, opened 2022-02-15
CCDB: Move, restyle, and reorder trends/list/map tabs

anselmbradford:
## Changes

- Moved TabbedNavigation to under the action bar in the Trends|List|MapPanel components.
- Give the active tab a white background and the inactive tabs a gray background to match their new location.
- Add icons to each tab to give a little hint about the shape of the data in each tab.
- Reorder the tabs to be trends, list, map
- Make trends the default tab so that it's active if you go to just consumerfinance.gov/data-research/consumer-complaints/search/

## Testing

- Pull branch and visit CCDB and click between tabs.

## Screenshots

Screen Shot 2022-02-15 at 11 24 49 AM

Screen Shot 2022-02-15 at 11 24 57 AM

Screen Shot 2022-02-15 at 11 25 02 AM

### DIFF
--- a/src/reducers/__tests__/query.spec.jsx
+++ b/src/reducers/__tests__/query.spec.jsx
@@ -373,12 +373,11 @@ describe( 'reducer:query', () => {
         from: 0,
         page: 1,
         searchAfter: '',
-        dataNormalization: undefined,
-        enablePer1000: true,
-        focus: '',
-        mapWarningEnabled: true,
-        tab: 'Map',
-        queryString: ''
+        chartType: "line",
+        focus: 'Yoyo',
+        tab: 'Trends',
+        queryString: '?focus=Yoyo',
+        trendsDateWarningEnabled: false
       } )
     } )
 
--- a/src/utils/reducers.jsx
+++ b/src/utils/reducers.jsx
@@ -46,8 +46,8 @@ export const enforceValues = ( value, field ) => {
       values: Object.keys( types.sorts )
     },
     tab: {
-      defaultVal: types.MODE_MAP,
-      values: [ types.MODE_MAP, types.MODE_TRENDS, types.MODE_LIST ]
+      defaultVal: types.MODE_TRENDS,
+      values: [ types.MODE_TRENDS, types.MODE_LIST, types.MODE_MAP ]
     }
   }
   if ( valMap[field] ) {


src/reducers/__tests__/query.spec.jsx
```diff
@@ -28,15 +28,16 @@ describe( 'reducer:query', () => {
...
+          '&lens=overview&trend_depth=5&trend_interval=month',
         searchText: '',
         searchField: 'all',
         page: 1,
         searchAfter: '',
         size: 25,
         sort: 'created_date_desc',
         subLens: '',
-        tab: types.MODE_MAP,
+        tab: 'Trends',
```
flacoman91, line 40: i think we have [types.MODE_TRENDS](https://github.com/cfpb/ccdb5-ui/blob/main/src/constants/index.jsx#L15)

src/reducers/__tests__/query.spec.jsx
```diff
@@ -372,6 +373,7 @@ describe( 'reducer:query', () => {
         from: 0,
         page: 1,
         searchAfter: '',
+        dataNormalization: undefined,
```
flacoman91, line 376: I dont think this test needs to be updated.

this line [here](https://github.com/cfpb/ccdb5-ui/blob/main/src/utils/reducers.jsx#L49) should be changed to 
      defaultVal: types.MODE_TRENDS

so that if we accidentally pass in an invalid tab name (coming from url or something, it makes the default to be the Trends tab.

src/reducers/__tests__/query.spec.jsx
anselmbradford: Hmm am I missing how this particular test is set up. I noticed in tests like https://github.com/cfpb/ccdb5-ui/blob/main/src/components/__tests__/RowChart.spec.jsx#L426 the literal value is used. Some schools of thought would say not to test against variable from the codebase in the test, because if the variable erroneously changed, the test would still pass. Which way makes sense here?

src/reducers/__tests__/query.spec.jsx
anselmbradford: Updated in https://github.com/cfpb/ccdb5-ui/pull/386/commits/1dc07cbb61cc6f8c1ddd316ff792c31d0f122911

anselmbradford: The test is updated to that of the Trends signature. That makes sense, right?

src/reducers/__tests__/query.spec.jsx
anselmbradford: Another example is `Trends` here --> https://github.com/cfpb/ccdb5-ui/blob/main/src/reducers/__tests__/query.spec.jsx#L384, while types.MODE_MAP is used here --> https://github.com/cfpb/ccdb5-ui/blob/main/src/reducers/__tests__/query.spec.jsx#L399

src/reducers/__tests__/query.spec.jsx
flacoman91: Yes it does. its that we want  it to be on the trends tab as defaultl.

src/reducers/__tests__/query.spec.jsx
flacoman91: I think the latter.   Doesn't really matter I think.  The values are pretty tightly coupled, but you're right, we should be consistent about how we do it.

flacoman91 (approved):
👍

PR #390, opened 2022-02-18
CCDB: Add narrative filter buttons to the list panel's refine bar

anselmbradford:
## Additions

- Adds narrative toggle buttons to the refine bar on the list view.

## Changes

- Moves the per capita styles to the refine bar styles to share them between the narrative and per capita refine bar buttons.

## Testing

- `yarn run test` should pass.
- Build and run locally and click the List view. 
  - Click the "Only complaints with narratives" button in the refine bar and scroll down to see that the narratives filter is checked.
  - Click the "All complaints" button in the refine bar and scroll down to see that the narratives filter is not checked.

## Screenshots

Screen Shot 2022-02-18 at 12 46 11 PM

### DIFF
--- a/src/actions/__tests__/filter.spec.jsx
+++ b/src/actions/__tests__/filter.spec.jsx
@@ -69,6 +69,20 @@ describe('action:filterActions', () => {
     })
   })
 
+  describe('addFilter', () => {
+    it('creates a simple action', () => {
+        const filterName = 'timely'
+        const filterValue = 'Yes'
+        const expectedAction = {
+          type: sut.FILTER_ADDED,
+          filterName,
+          filterValue,
+          requery: REQUERY_ALWAYS
+        }
+        expect(sut.addFilter(filterName, filterValue)).toEqual( expectedAction );
+    })
+  })
+
   describe('removeFilter', () => {
     it('creates a simple action', () => {
         const filterName = 'timely'


src/reducers/query.jsx
```diff
@@ -562,6 +562,29 @@ export function removeAllFilters( state ) {
...
 }
 
+/**
+* Adds a filter to the current set
+*
+* @param {object} state the current state in the Redux store
+* @param {object} action the payload containing the filter to add
+* @returns {object} the new state for the Redux store
+*/
+function addFilter( state, action ) {
```
flacoman91, line 572: we should add a test for this action for the 3 different paths you can take

- has_narrative
- existing filter
- no filter existing yet

flacoman91: Tests have been added.

src/actions/filter.jsx
```diff
@@ -93,6 +94,22 @@ export function toggleFlagFilter( filterName ) {
...
 }
 
+/**
+* Notifies the application that a filter was added
+*
+* @param {string} filterName which filter was clicked
+* @param {string} filterValue the value of the filter that was clicked
+* @returns {string} a packaged payload to be used by Redux reducers
+*/
+export function addFilter( filterName, filterValue ) {
```
flacoman91, line 104: A little nitpicky, but we have unit tests for these action creators too.   It's a pretty pointless test since it's just testing redux, and these tests would go away if we migrated to redux-toolkit and used slices which combines the action creators and reducers into a single file.

src/components/RefineBar/PerCapita.jsx
```diff
@@ -1,4 +1,3 @@
-import './PerCapita.less'
```
flacoman91, line 1: This would be a good candidate to also convert to a functional component like the narratives button

src/actions/filter.jsx
anselmbradford: I wasn't totally certain what action you were saying to do here (leave as-is, add addFilter test, or remove removeFilter test), but I've added the test in https://github.com/cfpb/ccdb5-ui/pull/390/commits/2553103062337dfb72088c222aedd67cfbf38753 and we can trim the superfluous ones altogether when you're back.

flacoman91:
Add an addFilter test.

These tests just test redux and really aren't useful in reality, but I have
caught some bugs since it forces me to re-read code

flacoman91 (approved):
Looks good!

src/components/RefineBar/PerCapita.jsx
anselmbradford: I'll get this in another PR

anselmbradford: https://github.com/cfpb/ccdb5-ui/pull/398

PR #391, opened 2022-02-18
Update modal z-index

anselmbradford:
Modal was appearing behind the menu because of the z-index

## Changes

- Add `z-index:100` to modal overlay.

## Testing

- No visual change inside ccdb running on its own. Once brought into cfgov, clicking the "Things to know before you use this database" modal will open a modal that covers the mega menu of cfgov.


higs4281 (approved):
Looks good locally and on DEV4.

PR #393, opened 2022-02-22
Adjust notification txt

anselmbradford:
## Changes

- Adjust notification txt

## Testing

- Change the chart options till there is no data to display (select a too small date range).


flacoman91 (approved):
Easy peasy

PR #394, opened 2022-02-23
Update TabbedNavigation.less to remove gap below inactive tab buttons

anselmbradford:
Reformat tabs CSS to eliminate gap below the tabs.

## Changes

- Convert em and `larger` font size keyword to `unit` calculation.
- Add after pseudoelement to hide bottom border of active button.
- Remove line-height adjustment.

## Testing

- Gap below the inactive tabs should be gone.

## Screenshots

Before:
Screen Shot 2022-02-23 at 11 00 10 AM

After:
Screen Shot 2022-02-23 at 11 00 18 AM

### DIFF
--- a/src/components/TabbedNavigation.less
+++ b/src/components/TabbedNavigation.less
@@ -6,6 +6,12 @@
   section {
     padding-left: 10px;
     .tab {
+      margin: 0;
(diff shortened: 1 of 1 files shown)


niqjohnson (approved):
Tabs look great locally

PR #397, opened 2022-02-24
Fix notification gap

anselmbradford:
Notification gap was added at https://github.com/cfpb/ccdb5-ui/commit/a71bbd883a6b53c81bf05f39c98a42a20190be79#diff-8406d995f97ed17de9731eb61b5e24539f22898742d518ce99661c54380a59f6R37, however, this was only for charts that appear inside the `external-tooltip` class, which doesn't appear on the Overview drop-down (only Company and Product).

## Changes

- Add gap between notification and note for all lenses.

## Testing

- Change filters so that a chart is not displayed and the notification and note show up. Change the lens drop-down and see that the gap between the notification and note is consistent.

## Screenshots

Before:
Screen Shot 2022-02-24 at 11 45 08 AM

After:
Screen Shot 2022-02-24 at 12 28 21 PM

### DIFF
--- a/src/components/Trends/TrendsPanel.less
+++ b/src/components/Trends/TrendsPanel.less
@@ -27,15 +27,15 @@
   }
   .chart {
     width: 100%;
+
+    .m-notification {
+      margin-bottom: 10px;
+    }
   }
   &.external-tooltip {
     section {
       &.chart {
         width: 70%;
-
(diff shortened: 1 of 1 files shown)


higs4281 (approved):
All results show up for me as shown.

PR #401, opened 2022-02-28
Ans fix area chart

imuchnik:
Moving this PR to the correct branch


flacoman91 (approved):
Just take all of the changes from ans fix area chart in the **dist** conflicts.

PR #404, opened 2022-03-04
Provide overrides to fix legacy CF rules

anselmbradford:
## Changes

- Provide overrides to fix legacy CF rules

## Testing

- Run cfgov locally.
- In this branch, run `yarn build`
- Copy and rename `/dist/ccdb5.min.js` to `cfgov/static_built/apps/ccdb-search/js/main.js`
- Visit http://localhost:8000/data-research/consumer-complaints/search/ and compare header and footer to production header and footer.


flacoman91 (approved):
👍

PR #408, opened 2022-08-18
Reverses Create React App eject | replaces Capital Framework with the Design System

flacoman91:
This PR reverses the `react-scripts eject` from create react app and implements the linting rules and utils we have in an attempt to standardize the code bases from Complaint Explorer and CCDB.

Since the build command now generates a separate css and js file, we can get this to continue to work in CFGOV by updating the [main.less](https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/apps/ccdb-search/css/main.less) by adding `@import '../node_modules/@cfpb/ccdb5-ui/dist/ccdb5.css';`  to the top of the file.

**see example**
https://github.com/cfpb/consumerfinance.gov/pull/7283/files

## Changes

- `xit` some unit tests that need to be refactored.
- Adds fonts.sh script for adding webfonts required by the build.
- Add craco config and remove unused dependencies.
- Replace capital framework with design system and update expandables markup.

## Testing

- `yarn && yarn start` and visit http://localhost:3000/ and see that app still works. 

## Screenshots

Note that this bug is showing up and will need to be fixed in the DS https://github.com/cfpb/design-system/issues/1398

Screen Shot 2022-10-04 at 12 42 49 PM


JeffreyMFarley:
Hooray!

KatherynRizek:
@anselmbradford I'm getting some snapshot failures. Should I see what happens when I update them? A lot look like the date got shifted:

Screen Shot 2022-10-05 at 9 46 12 AM
Screen Shot 2022-10-05 at 9 47 25 AM

src/__tests__/App.spec.js
```diff
@@ -59,7 +59,7 @@ describe('initial state', () => {
     );
 
     expect(
-      screen.getByRole('button', { name: /Back to search results/ })
+      screen.getByRole('button', { name: /Go to search home page/ })
```
KatherynRizek, line 62: Shouldn't this be the detail view that's rendered, which would have `Back to search results` in it?

anselmbradford: @flacoman91 want to fill in here?

src/setupTests.js
```diff
@@ -1,29 +1,9 @@
...
-Object.defineProperty(document, 'referrer', {
-  configurable: true,
-  value: 'http://www.example.org',
-});
-
-// Date and time (EDT): Tuesday, May 5, 2020 12:00:00 AM
-window.MAX_DATE = Date.UTC(2020, 4, 5, 4);
-
-MockDate.set('5/5/2020');
+jest.useFakeTimers().setSystemTime(new Date('2020-05-05'));
```
KatherynRizek, line 9: This might be related to the date mismatch

src/__tests__/App.spec.js
flacoman91: Yes, ignore this.  The way this works is a bit problematic and is being worked on here:
https://github.com/cfpb/ccdb5-ui/pull/436

anselmbradford:
> I'm getting some snapshot failures. Should I see what happens when I update them? A lot look like the date got shifted:

Ughh so I was running into that and to get both my local and GH actions to pass I had to adjust the date in `jest.useFakeTimers().setSystemTime(new Date('2020-05-05'));` in `setupTests.js` so that it was one day ahead `2020-05-06`, generate new snapshots, but not commit the change to `setupTests.js`. I thought since the snapshots will be going away it wasn't worth debugging, but now I see the change to `setupTests.js` needs to be there but uncommitted for both to pass. Any ideas? I have a branch that strips out the snapshots if that's easier.

KatherynRizek:
> > I'm getting some snapshot failures. Should I see what happens when I update them? A lot look like the date got shifted:
> 
> Ughh so I was running into that and to get both my local and GH actions to pass I had to adjust the date in `jest.useFakeTimers().setSystemTime(new Date('2020-05-05'));` in `setupTests.js` so that it was one day ahead `2020-05-06`, generate new snapshots, but not commit the change to `setupTests.js`. I thought since the snapshots will be going away it wasn't worth debugging, but now I see the change to `setupTests.js` needs to be there but uncommitted for both to pass. Any ideas? I have a branch that strips out the snapshots if that's easier.

They seem to be passing now. What was the fix?

anselmbradford:
> They seem to be passing now. What was the fix?

I bumped the date in setupTests https://github.com/cfpb/ccdb5-ui/pull/408/files#diff-f1995b7576ce098f06844c16df12561f4694c56c84869ec5180933e652f56862R9

Buuuuut… GH Actions were down today, so they likely won't pass there.

JJZolper (approved):

PR #409, opened 2022-08-18
Add tour to ccdb

flacoman91:

## Additions

-

## Removals

-

## Changes

-

## Testing

-

## Notes

-

## Todos

-

## Checklist


flacoman91:
@anselmbradford this is almost ready.  I still have some things I need to work out in the Tour component.

Could you take over and assist with the styling?

src/actions/trends.jsx
```diff
@@ -107,6 +108,21 @@ export function removeFocus() {
   }
 }
 
+/**
+ * Notifies the application that the tour was toggled
+ *
+ * @param {string} value the new payload from the tooltip
+ * @returns {string} a packaged payload to be used by Redux reducers
+ */
+export function tourToggled( value ) {
```
flacoman91, line 117: I would write a test for this, but it's unnecessary esp if we are going to convert to redux toolkit.

src/components/Tour/Tour.jsx
```diff
@@ -0,0 +1,111 @@
...
+   * Before Change handler
+   *
+   * @param {object} ref - React component reference.
+   */
+  function handleBeforeChange( ref ) {
+    const expand = ref.current.introJs.currentStep() + 1;
+
+    // hacks to setup the page. we should figure out if we should
+    // have this in redux instead
+    if ( expand === 9 ) {
```
anselmbradford, line 46: Should the `9` go into a constant?

src/components/Tour/Tour.jsx
```diff
@@ -0,0 +1,111 @@
...
+      // search tips open/closed state should be moved to redux
+      // and when the tour start we should "reset" the page
+      document.querySelector( '.advanced-container button' ).click()
+    }
+
+    if ( expand === 16 ) {
```
anselmbradford, line 52: Should the `16` go into a constant?

src/components/Tour/Tour.jsx
```diff
@@ -0,0 +1,111 @@
...
+      // this is a hacky solution for the Trends Row chart.
+      // we need to figure out how many expandables there are
+      // figure out how many ticks there are.
+      // if not equal, it means one is open so we don't need to expand a tick
+      // i would prefer to collapse all of the bars and expand only the first.
+      // this way you can actually highlight the "View more about XYZ Product"
+      const expandables = document.querySelectorAll( '#row-chart-product .y-axis-group .tick.expandable' );
+      const ticks = document.querySelectorAll( '#row-chart-product .y-axis-group .tick' );
```
anselmbradford, line 60: You could eliminate the multiple row-chart-product ID lookups:

```js

const rowChartProduct = document.querySelector( '#row-chart-product' );
const expandables = rowChartProduct.querySelectorAll( '.y-axis-group .tick.expandable' );
const ticks = rowChartProduct.querySelectorAll( '.y-axis-group .tick' );

```

anselmbradford: Also, if there's only one `y-axis-group`, you could include that in the first query.

src/components/Tour/TourButton.less
```diff
@@ -0,0 +1,11 @@
+@import (less) "../../css/base.less";
```
anselmbradford, line 1: This looks like it could be `@import (reference) "../../css/base.less";`

src/components/Tour/Tour.jsx
flacoman91: I updated this so it's more scalable so we don't have to hardcode the number.

src/components/Tour/Tour.jsx
flacoman91: same as above.

src/components/Tour/Tour.jsx
flacoman91: i moved this logic into redux so the page state gets set to a good happy path.  We don't need to do this check anymore.

src/components/Tour/Tour.less
```diff
@@ -0,0 +1,16 @@
+// Intro.js styles
+@import (css) '../../../node_modules/intro.js/introjs.css';
```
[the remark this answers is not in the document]
flacoman91: No.  This may be part of the technical debt regarding our build tools. 

Screen Shot 2022-08-22 at 6 08 37 PM

src/components/Tour/TourButton.less
flacoman91: this one looks okay.

flacoman91:
Looks like  the cookie test  all of a sudden failing is a  known  issue

https://github.com/salesforce/tough-cookie/issues/246

anselmbradford:
I'm getting an error on the end tour step

 
Screen Shot 2022-09-01 at 12 29 58 PM

flacoman91:
updates responding to GHE  2104 have been  made.

flacoman91:
removed Date Range steps in trends and map as requested.

flacoman91:
Restored links and table of contents. made only "Index" a title heading

Screen Shot 2022-09-07 at 3 39 36 PM
Screen Shot 2022-09-07 at 3 39 04 PM

src/components/RefineBar/Select.jsx
```diff
@@ -48,7 +48,7 @@ export class Select extends React.Component {
     const values = this.getValues()
 
     return (
-      <section className="cf-select">
+      <section className={'cf-select ' + this.props.id}>
```
anselmbradford, line 51: Should `this.props.id` go in an `id` attribute instead of class?

package.json
```diff
@@ -149,5 +149,9 @@
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org/"
+  },
+  "dependencies": {
```
anselmbradford, line 153: I think we want everything on `devDependencies`, as I think `dependencies` end up in the npm package. Since the npm package is just delivering a built JS file, I don't think we want anything else there.

src/components/RefineBar/Select.jsx
anselmbradford: Ah I see in the snapshot it gets repeated. Well, instead of super generic class names like `size` and `sort`, let's prefix this to be `<section className={'cf-select cf-select__' + this.props.id}>`

flacoman91: I had to add this here so that I could target these for highlighting in intro.js here

https://github.com/cfpb/ccdb5-ui/pull/409/files#diff-1f017f3f6619b22aabaa7855af7aa91b94892609d8f7f765b5d5c942c4565d98R60

I think it might be possible to update the document.querySelector call to look for attributes instead

src/components/Tour/constants/tourStepsConstants.js
```diff
@@ -0,0 +1,370 @@
...
+    {
+      element: DOCUMENT.STEP_4,
+      intro: 'This link will take you back to the search results'
+    }
+  ],
+  [MODE_MAP]: [
+    {
+      element: GEO.STEP_1.selector,
+      intro:
+         '<div>' +
```
anselmbradford, line 69: trivial: indented a little too much.

src/components/RefineBar/Select.jsx
anselmbradford: Hm yea, that actually would be the best way. The prefixed class sets this up as a modifier on `cf-select`, which implies that `cf-select__sort` (or whatever) is modifying `cf-select`. In actuality it is being used as an ID. That's not a big deal, but it could send someone off to look for the modifying CSS. If you want, you should be able to do `<section className="cf-select" data-tour={this.props.id}>` and query it with `document.querySelector('[data-tour="sort"]')`, or `document.querySelector('section[data-tour="sort"]')`.

anselmbradford (approved):
You could update the data- attributes, but otherwise looks good

src/components/RefineBar/Select.jsx
```diff
@@ -44,19 +44,21 @@ export class Select extends React.Component {
   }
 
   render() {
-    const id = 'choose-' + this.props.id
+    const id = 'select-' + this.props.id
```
flacoman91, line 47: i just renamed this to select vs choose, since that makes more sense.

flacoman91:
> You could update to use data- attributes, but otherwise looks good

I updated it [here](https://github.com/cfpb/ccdb5-ui/pull/409/files#diff-1ea787e5c97d51c9f0a9bdf61d2fc333eeea3f1dad7394c10c1dc8bd61f1d568R52) and also renamed the id to make it clearer.

PR #414, opened 2022-08-23
1764 - migrate momentjs -> dayjs

flacoman91:
reduce techical debt. get off momentjs since its deprecated and dayjs  is better. 

Fixed unit tests
added some validation to convert dashes to /

### DIFF
--- a/src/utils/__tests__/chart.spec.jsx
+++ b/src/utils/__tests__/chart.spec.jsx
@@ -71,10 +71,9 @@ describe( 'getTooltipTitle', () => {
     interval = 'Month'
     const inDate = '2015-09-01T10:00:00.000Z'
     res = sut.getTooltipTitle( inDate, interval, dateRange, true )
-    // this is wrong in tests, but correct in the UI.
-    expect( res ).toEqual( 'Date range: 9/1/2015 - 10/31/2015' )
+    expect( res ).toEqual( 'Date range: 9/1/2015 - 9/30/2015' )
     res = sut.getTooltipTitle( inDate, interval, dateRange, false )
-    expect( res ).toEqual( '9/1/2015 - 10/31/2015' )
+    expect( res ).toEqual( '9/1/2015 - 9/30/2015' )
   } )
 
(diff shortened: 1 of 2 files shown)


src/constants/index.jsx
```diff
@@ -4,6 +4,9 @@ export const MODAL_SHOWN = 'MODAL_SHOWN'
 
 export const DATE_RANGE_MIN = '2011-12-01T12:00:00.000Z'
 
+export const DATE_VALIDATION_FORMAT = 'M/D/YYYY';
```
anselmbradford, line 7: You could add a comment above this line with a link to `https://day.js.org/docs/en/display/format` so it's clearer this is set by dayjs.

src/utils/__tests__/chart.spec.jsx
```diff
@@ -129,7 +133,8 @@ describe( 'getTooltipTitle', () => {
     const inDate = '2020-07-01T00:00:00.000Z'
     res = sut.getTooltipTitle( inDate, interval, dateRange, true )
     expect( res ).toContain( 'Date range: 7/14/2020 - ' )
-    expect( res ).toEqual( 'Date range: 7/14/2020 - 9/30/2020' )
+    // TODO: figure out why this doesn't work in CI
+    // expect( res ).toEqual( 'Date range: 7/14/2020 - 9/30/2020' )
```
anselmbradford, line 137: You want to do this in another PR or this one?

flacoman91: I'd do this in another one. as far as i  could tell, I  could never get this to work in CI even when it was using momentjs.

flacoman91: Sorry, i took another look at this.   Looks like some of the other cases  were  incorrect and i fixed up those   tests and updated the logic.

flacoman91: and it passes locally and in CI now!

anselmbradford (approved):

PR #423, opened 2022-09-20
Add PropType validation and update boolean variable names

anselmbradford:
## Changes

- Change boolean variable names to use `is`/`has` naming.
- Adds PropType validations to components.

## Testing

- `yarn test` should pass.


flacoman91 (approved):
👍  great

PR #424, opened 2022-09-21
2262 - Update eslint config

anselmbradford:
Addresses: https://github.cfpb.gov/data-products-team/explorer-docs/issues/2262

## Changes

- Update eslint config to match explorer.
- Fix linting errors.
- Add cypress linting command

## Testing

- `yarn lint` should show no errors (it will show warnings).

## Todos

- The a11y eslint plugin is temporarily disabled, as the fixes there are more involved.
- The cypress linting command still shows errors. We'll want to fix those and then add it to the main linting command (as is done in explorer)


flacoman91 (approved):
👍   yay to standards

PR #428, opened 2022-09-23
Fix tooltips date range

KatherynRizek:

## Additions

- Make input date start at beginning of the day for consistency 
- In the month switch statement, add a month before subtracting a day

## Testing

- Current tests pass; they were previously passing due to some funkiness with the timezones. Adding the extra month broke them but making the input date start at the beginning of the day took away the timezone discrepancies. 

## Screenshots
Screen Shot 2022-09-23 at 12 20 44 PM

## Checklist

### DIFF
--- a/src/utils/__tests__/chart.spec.js
+++ b/src/utils/__tests__/chart.spec.js
@@ -69,8 +69,7 @@ describe('getTooltipTitle', () => {
 
   it('sets tooltip title - month', () => {
     interval = 'Month';
-    let inDate = new Date('September 1, 2015 UTC');
-    inDate = inDate.toISOString();
+    const inDate = '2015-09-01T10:00:00.000Z';
     res = sut.getTooltipTitle(inDate, interval, dateRange, true);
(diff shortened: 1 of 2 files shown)


KatherynRizek:
Hmm... Looking at why that unit test is failing here when it passes locally

flacoman91:
> Hmm... Looking at why that unit test is failing here when it passes locally

I've encountered that issue before and it's a tricky one.   TZ in CI is probably different from what is running locally.  
I would copy as much of the logic from explorer as possible since they should be very similar, if not the same thing.

We've had to set the tz here before:
https://github.com/cfpb/ccdb5-ui/blob/main/src/utils/chart.js#L16

KatherynRizek:
The tests finally pass! And the fix still seems to be working:
Month:
Screen Shot 2022-09-26 at 1 07 41 PM

Quarter:
Screen Shot 2022-09-26 at 1 07 27 PM

anselmbradford (approved):

PR #429, opened 2022-09-23
Remove unreferenced packages

anselmbradford:
Seems like we had a lot of dependencies that aren't actually used. This PR removes them.

## Removals

- Remove unreferenced packages from `package.json`.

## Testing

- PR checks should pass.


KatherynRizek (approved):
Everything starts and tests pass. I also double checked that the removed packages aren't referenced. Looks good 👍

PR #434, opened 2022-09-30
list to function components

KatherynRizek:

## Additions

- Add react testing library tests for ComplaintCard, ListPanel and Pagination
- Add redux toolkit
- Add folders to keep source files, test files and styling files together

## Removals

- Remove old tests for ComplaintCard, ListPanel and Pagination

## Changes

- Downgrade react testing library to be compatible with React 16
- Convert ComplaintCard, ListPanel and Pagination to be functional components using `useDispatch` and `useSelector` instead of `mapStateToProps` and `mapDispatchToProps`

## Testing

- Make sure the list view still renders complaints
- Ensure you can still change the pages
- Run test suite

## Notes

-

## Todos

-

## Checklist

### DIFF
--- a/src/components/List/ListPanel/ListPanel.less
+++ b/src/components/List/ListPanel/ListPanel.less
@@ -1,4 +1,4 @@
-@import (less) '../../../css//base.less';
+@import (less) '../../../css/base.less';
 
 .list-panel {
   // The "No results were found for your search" heading.


src/components/List/ComplaintCard/ComplaintCard.spec.js
```diff
@@ -0,0 +1,127 @@
...
+
+  test('it renders correctly', () => {
+    const tree = renderer
+      .create(
+        <IntlProvider locale="en">
+          <ComplaintCard row={itemFixture} />
+        </IntlProvider>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
```
KatherynRizek, line 37: I could add this snapshot in easily enough since the component doesn't depend on the redux store but I didn't for other two tests. Do we still want to be snapshot testing as we upgrade the tests from Enzyme to Testing Library? I can look into how to do it with Redux if we do; just wanted to check in about it.

anselmbradford:
What node are you on? I'm getting some errors about incompatible node modules that are requiring node 18. I'm on 17. Not sure we want to go to 18 yet?

KatherynRizek:
> What node are you on? I'm getting some errors about incompatible node modules that are requiring node 18. I'm on 17. Not sure we want to go to 18 yet?

I think I'm on v16.17.1

src/components/List/ComplaintCard/ComplaintCard.spec.js
flacoman91: I'm okay with not doing snapshot testing and removing them from CCDB as long as we're testing for the values we need via testing-library.

package.json
```diff
@@ -139,5 +139,9 @@
   "publishConfig": {
     "access": "public",
     "registry": "https://registry.npmjs.org/"
+  },
+  "dependencies": {
+    "@reduxjs/toolkit": "^1.8.5",
+    "@testing-library/jest-dom": "^5.16.5"
```
flacoman91, line 145: I did this too. 

all of the dependencies should be dev dependencies so cf.gov doesn't install anything we're using and only takes the minified bundles we generate.

flacoman91:
> * Downgrade react testing library to be compatible with React 16

Thanks for doing this and keeping me honest. 
I was going to be lazy/procrastinate, comment out any tests I can and then wait to upgrade to React 18 first.

package.json
KatherynRizek: Oh, good to know. I'll fix that

src/testUtils/test-utils.js
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ *
+ * @param {object} ui - React component we are passing in
+ * @param {object} root0 - React component
+ * @param {object} root0.preloadedState - The initial component state
+ * @param {object} root0.store - Redux store we are mocking
+ * @returns {object} React jsx components
+ */
+function testRender(
```
flacoman91, line 24: 👏

package.json
KatherynRizek: Updated in 07108f0

src/components/List/ComplaintCard/ComplaintCard.spec.js
KatherynRizek: Removed the ComplaintCard snapshot and added additional assertions in 2a656e9

src/components/List/ListPanel/ListPanel.less
```diff
@@ -1,4 +1,4 @@
-@import (less) '../../css/base.less';
+@import (less) '../../../css//base.less';
```
flacoman91, line 1: i think the double slash is a typo here

KatherynRizek: Yep, thanks for catching that. Fix is pushed

src/components/List/ListPanel/ListPanel.spec.js
```diff
@@ -0,0 +1,178 @@
...
+    };
+    const newResultsState = { items: [itemFixture] };
+
+    renderComponent(newAggsState, defaultQuery, newResultsState, defaultView);
+
+    expect(screen.getByText('JP Morgan')).toBeDefined();
+    expect(screen.getByText('11/16/2022')).toBeDefined();
+  });
+
+  test('onSize triggers dispatch and analtyics event', () => {
```
flacoman91, line 113: this is so much better and more thorough than the previous method of testing mapDispatchtoProps

🍾

PR #436, opened 2022-10-05
Complaint Detail view functional component conversion

flacoman91:
This pr allows us to leverages more from React Router 6 and some refactoring

1. Used Link component from react-router on list view
2. Better logic for rendering template
3. convert to functional component
4. Bug fixes for Back button in Complaint detail view
5. Added a test renderer with MemoryRouter that allows us to test and use the useLocation and useParams hooks
6. Removes the window.location calls from Complaint detail view in favor of hooks.  It also makes it way easier to test.

To do: 
- rewrite tests for ResultsPanel.js.  I had to tweak the test for list view since switching to the Link Component caused the react-test-renderer to break.  
- convert hero to functional component and change data link to use relative link from constant defined in this PR.

### DIFF
--- a/src/components/ComplaintDetail/ComplaintDetail.spec.js
+++ b/src/components/ComplaintDetail/ComplaintDetail.spec.js
@@ -1,4 +1,4 @@
-import { ComplaintDetail } from '../ComplaintDetail/ComplaintDetail';
+import { ComplaintDetail } from './ComplaintDetail';
 import React from 'react';
 import {
   testRenderWithMemoryRouter as render,
--- a/src/reducers/detail/detail.spec.js
+++ b/src/reducers/detail/detail.spec.js
@@ -1,5 +1,5 @@
 import * as sut from '../../actions/complaints';
-import target from '../detail/detail';
+import target from './detail';
 
 describe('reducer::detail', () => {
   it('has a default state', () => {


anselmbradford:
Not a big deal as they can be generated before release, but it looks like you could run `yarn build` to get the dist files updated. It looks like they're being deleted right now, without re-generation.

flacoman91:
> 

I actually deleted them because when you run yarn build, it looks like it's a different file name without the min.js

flacoman91:
I just added them.  For some reason it was showing up as untracked in git.

dist/ccdb5.js.LICENSE.txt
```diff
@@ -47,6 +47,17 @@ object-assign
...
+ * React Router DOM v6.4.1
+ *
+ * Copyright (c) Remix Software Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE.md file in the root directory of this source tree.
+ *
+ * @license MIT
+ */
+
```
anselmbradford, line 60: Weird, this dupe is back again.

src/components/ComplaintDetail/ComplaintDetail.js
```diff
@@ -0,0 +1,62 @@
...
+  const dispatch = useDispatch();
+  const data = useSelector(selectDetailData);
+  const error = useSelector(selectDetailError);
+  const activeCall = useSelector(selectDetailActiveCall);
+  const search = useSelector(selectQuerySearch);
+
+  const isLoading = activeCall !== '';
+  const backUrl = useMemo(() => {
+    const pathName = location.pathname || '';
+    const idx = pathName.indexOf('detail');
```
anselmbradford, line 28: If `pathName` ends up as '', `idx` will end up as `''.indexOf('detail')`, which will equal `-1`. Is it a problem if `idx` equals `-1`?

anselmbradford: Or, if `pathName` would never end up in that state, we should strip `|| ''` from the prior line.

flacoman91: The || '' is for when the page first loads that the object isn't ready it.

It seemed to bomb on initial load.

I guess I can just return an empty string and exit if location isn't ready yet.

src/components/ComplaintDetail/ComplaintDetail.js
```diff
@@ -0,0 +1,62 @@
...
+      <nav className="layout-row">
+        <div className="back-to-search flex-fixed">
+          <Link to={backUrl}>
+            {iconMap.getIcon('left', 'cf-icon-left')}
+            Back to search results
+          </Link>
+        </div>
+        <div className="meaning flex-fixed">
+          <a
+            href="https://www.consumerfinance.gov/complaint/data-use/"
```
anselmbradford, line 47: We possibly should turn these into relative links when they go off to production. I.e. have a search and replace in `build.sh` that turns `https://www.consumerfinance.gov/` into `/`.

flacoman91: Good point. 

I was about to fix it in this component like so
```
  const dataUseLink =
    window.location.origin.indexOf('localhost') > -1
      ? 'https://www.consumerfinance.gov/complaint/data-use/'
      : '/complaint/data-use/';
```

I also checked to see where else this occurs. 

Looks like it's in the hero as well.
https://github.com/cfpb/ccdb5-ui/search?q=https%3A%2F%2Fwww.consumerfinance.gov

hero isn't a functional component yet and has a snapshot test that would break in CI so I won't touch it. 

I moved the link into a constant that we can reference when we convert hero and remove the snapshot. 
updated in [1e7ff25f862eb59aba9b54e128d62b9bfebaa18e](https://github.com/cfpb/ccdb5-ui/pull/436/commits/1e7ff25f862eb59aba9b54e128d62b9bfebaa18e)

KatherynRizek (approved):
Looks good 👍

PR #439, opened 2022-10-07
adding class to tour button for tracking

flacoman91:
Adding a-btn class so that GA can target the button click.


src/components/Tour/TourButton.js
```diff
@@ -8,7 +8,10 @@ export const TourButton = () => {
   const dispatch = useDispatch();
 
   return (
-    <button onClick={() => dispatch(tourShown())} className="tour-button">
+    <button
+      onClick={() => dispatch(tourShown())}
+      className="tour-button a-btn"
```
anselmbradford, line 13: Trivial, but I would swap these so it's `a-btn tour-button` so it goes general -> specific. `a-btn` is a widely used class across the codebase and `tour-button` is specific to this use-case. If we ever made `tour-button` a modifier on `a-btn`, it would be `a-btn a-btn__tour` or whatever, so it would already be set up for that.

src/setupTests.js
```diff
@@ -6,4 +6,4 @@ import '@testing-library/jest-dom';
 
 configure({ adapter: new Adapter() });
 
-jest.useFakeTimers().setSystemTime(new Date('2020-05-06'));
+jest.useFakeTimers().setSystemTime(new Date('2020-05-05T04:00:00.000Z'));
```
anselmbradford, line 9: Oh nice, is this the solution???

anselmbradford (approved):

src/setupTests.js
flacoman91: Yeah, I think the lack if offset hours tz causes the issue.   This way we explicitly set the time including day.

PR #440, opened 2022-10-12
Update DS to 0.17.11 to bring in expandable updates

anselmbradford:
Update DS to 0.17.11 to bring in the fix to expandable headers so that the toggle doesn't shift down when closed.

## Changes

- Update DS to 0.17.11

## Testing

- `yarn && yarn start` and open and close the filters and see that the toggle doesn't shift down.

## Screenshots

Screen Shot 2022-10-12 at 12 24 45 PM


flacoman91 (approved):
lgtm.  I tested this in CFGOV

Screen Shot 2022-10-12 at 12 47 04 PM

maybe that CI test will pass if you run it again. 

 can you update the js and css bundles with a `yarn build` ?

anselmbradford:
> can you update the js and css bundles with a yarn build ?

Oops! Yup, updated

flacoman91:
cool.  I rebuilt and the files seem different, but i guess it doesnt matter since a new npm push would need a rebuild anyway.

PR #444, opened 2022-10-13
Refine bar to functional components

flacoman91:
Refactor RefineBar and child components correctly. 

- [x] ChartToggles
- [x] PerCapita
- [x] NarrativesButton
- [x] Select

## Additions
- added testing-library/user-event
- More consistency for buttons for Chart,  Per Capita,  Narratives buttons. When an option is selected, it gets a disabled state and users cannot click on it.  Mouseover state shows disabled.
- updated Cypress tests

## Removals
- Removed unnecessary redux action checks since users aren't able to click buttons when they are disabled. 
- fix/remove snapshot tests
- remove unnecessary disabled class

## Changes
- convert size param to string so Select dropdown can handle selected option validation correctly
- convert enzyme tests to use testing-library

## Testing
tests should still pass
- yarn install
- yarn test
- yarn cypress run

## Notes

-

## Todos

-

## Checklist


package.json
```diff
@@ -53,6 +53,7 @@
     "@reduxjs/toolkit": "^1.8.5",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "12.1.5",
+    "@testing-library/user-event": "12.1.5",
```
KatherynRizek, line 56: I think the newest version of user-event is on 14. 
The set up is a little different if you upgrade: https://testing-library.com/docs/user-event/setup

flacoman91: Yeah, I was trying to keep it on the same version of react testing-library. 

I know we downgraded testing-library since we're not on react 18 yet.

cypress/e2e/list/list.cy.js
```diff
@@ -89,27 +89,23 @@ describe('List View', () => {
...
-      cy.get(removeNarrativesButton).should('have.class', 'deselected');
 
       cy.get(filterHasNarrative).should('be.checked');
 
       // Click the all results button. The narratives should be removed.
       cy.get(removeNarrativesButton).click().should('have.class', 'selected');
-      cy.get(addNarrativesButton).should('have.class', 'deselected');
 
       cy.get(filterHasNarrative).should('not.be.checked');
     });
```
KatherynRizek, line 111: Line 175 is failing this test since it can't click on something disabled. Should we just remove `.click()`?

package.json
flacoman91: Looks like I was able to upgrade it here
https://github.com/cfpb/ccdb5-ui/pull/444/commits/994d6783b9e323e840d2f43efbb78d9742ca744b

There was some minor change but took some time to figure out the syntax.  We can update this in explorer too.

cypress/e2e/list/list.cy.js
flacoman91: I'm not seeing the failure, and it's passing in CI too.

KatherynRizek: Might be my local then. I'll go ahead and approve if you're passing and so is CI

KatherynRizek (approved):

PR #449, opened 2022-10-19
Fix font.sh curl commands

anselmbradford:
Oops the curl commands here for the fonts were incorrect and were downloading zero kb files!

## Changes

- Fixes curl command URLs and flags to be correct in fonts.sh script.

## Testing

- PR checks should pass.
- Running ./scripts/fonts.sh should produce font files that have a file size.


flacoman91 (approved):
tested and it works.

I think it would be nice to see an indication of the files installing successfully though, or even something like 

`ls src/static/fonts`  to confirm the files are there.

anselmbradford:
> I think it would be nice to see an indication of the files installing successfully though, or even something like

I removed the silent `s` flag.

PR #453, opened 2022-10-21
Dialogs to functional

flacoman91:

## Additions

- Better unit tests using RTL
- adding missing test coverage for tour and advanced search tips

## Removals

- Modal reducer - moved contents of modal reducer into View since there's only 1 value we use for it.
- showMoreAboutDialog action. This isn't necessary and just duplicates the showModal action
- Remove the download file link logic.  I think this may be some remnants of some IE11 things or code that wasn't supported when this code was written. Removed domUtils since we don't actually need to make a fake link in the dom to trigger a click anymore
- remove the `required` labels from Data Export form since these values are radio buttons and there's no way to have any empty radio buttons
- snapshot testing

## Changes

- Simplified the logic to show modals
- Moved AdvancedTips out of Dialogs since it isn't a modal component
- leverage use of modern react hooks to enable better form validation in the DataExport dialog
- Moved Loading out of Dialogs directory since it's not a modal component
- Moved Export Confirmation into its own component
- removed actions/dataExports into the DataExport components directory since the contents of this file are not redux actions. 

## Testing

- build code
- Click export data
- Download file as csv or json
- do the same again with filtered data
- confirm copy link works

## Notes

-

## Todos

-

## Checklist

### DIFF
--- a/src/components/Dialogs/DataExport/DataExport.spec.js
+++ b/src/components/Dialogs/DataExport/DataExport.spec.js
@@ -208,20 +208,23 @@ describe('DataExport', () => {
     expect(radioCsv).toBeChecked();
     expect(radioJson).not.toBeChecked();
 
+    expect(screen.getByRole('textbox')).toHaveValue(
+      'https://files.consumerfinance.gov/ccdb/complaints.csv.zip'
+    );
+
     fireEvent.click(radioJson);
 
+    await waitFor(() => {
+      expect(screen.getByRole('textbox')).toHaveValue(
+        'https://files.consumerfinance.gov/ccdb/complaints.json.zip'
+      );
+    });
     await waitFor(() => {
       expect(radioJson).toBeChecked();
     });
 
     await waitFor(() => {
       expect(radioCsv).not.toBeChecked();
-    });
-
-    await waitFor(() => {
-      expect(screen.getByRole('textbox')).toHaveValue(
-        'https://files.consumerfinance.gov/ccdb/complaints.json.zip'
-      );
     });
 
     fireEvent.click(radioCsv);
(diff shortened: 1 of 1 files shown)


src/components/Dialogs/DataExport/DataExport.spec.js
```diff
@@ -0,0 +1,256 @@
...
+
+    expect(
+      screen.getByRole('button', { name: 'Start export' })
+    ).toBeInTheDocument();
+    expect(screen.getByRole('button', { name: 'Cancel' })).toBeInTheDocument();
+    const buttonCopy = screen.getByRole('button', { name: 'Copy' });
+    expect(buttonCopy).toBeInTheDocument();
+    expect(buttonCopy).toHaveClass('a-btn__secondary');
+    fireEvent.click(buttonCopy);
+    await (() => {
```
KatherynRizek, line 63: This should be `await waitFor( () => {...} ); ` The `await` by itself will pass everything

src/components/Dialogs/RootModal.spec.js
```diff
@@ -0,0 +1,73 @@
...
+    //   .mockImplementation(() => jest.fn());
+    render(<RootModal />, {
+      preloadedState: {
+        view: {
+          modalTypeShown: types.MODAL_TYPE_EXPORT_CONFIRMATION,
+        },
+      },
+    });
+    expect(screen.getByText('Export complaints')).toBeInTheDocument();
+    // TODO: uncomment this out when we upgrade to react 18.
```
KatherynRizek, line 50: Hopefully the upgrade will fix this!

src/components/Dialogs/DataExport/DataExport.spec.js
flacoman91: awesome, thanks!

src/components/Dialogs/RootModal.spec.js
flacoman91: Looks like it did.  Good stuff 👍

src/components/Dialogs/DataExport/DataExport.spec.js
flacoman91: You're right.  By adding waitFor and also following the eslint nag of not combining the await conditions, an error was caught.

KatherynRizek (approved):

PR #457, opened 2022-10-27
CCDB - Tile Chart convert to functional components

flacoman91:
Converts TileMap to a functional component

## Additions

- better tests using RTL

## Removals

- snapshots

## Changes

- Moved Tile Map component to be under Map Directory to reflect site structure

## Testing

-

## Notes

-

## Todos

We're using user-event 14. Currently, we're calling userEvent.click, hover, etc directly instead of doing `await user.click` or   `await.hover` Best practice would be to do
`const user = userEvent.setup()` as shown here: https://testing-library.com/docs/user-event/setup

 but there's some more work to get that working.  See: https://github.com/testing-library/user-event/issues/1054

## Checklist

### DIFF
--- a/src/components/Map/TileChartMap/TileChartMap.js
+++ b/src/components/Map/TileChartMap/TileChartMap.js
@@ -31,7 +31,6 @@ export const TileChartMap = () => {
       o.abbr = o.name;
       o.fullName = stateInfo.name;
       o.perCapita = getPerCapita(o, stateInfo);
-      // o.className = getStateClass(stateFilters, o.name);
       return o;
     });
   }, [stateMapResultsState]);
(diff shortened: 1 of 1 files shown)


src/components/Map/TileChartMap/TileChartMap.spec.js
```diff
@@ -0,0 +1,254 @@
...
+
+    const newView = {
+      isPrintMode: false,
+      width: 1000,
+    };
+
+    renderComponent(newMap, {}, newView);
+    expect(screen.getByTestId('tile-chart-map')).toBeInTheDocument();
+    expect(screen.getByTestId('tile-chart-map')).not.toHaveClass('print');
+    await waitFor(() => {
```
KatherynRizek, line 69: I think it might be better practice to write it this way:
```
expect(await screen.findByText('FL')).toBeInTheDocument();
```
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#:~:text=Using%20waitFor%20to%20wait%20for%20elements%20that%20can%20be%20queried%20with%20find*

KatherynRizek: I stumbled across the article recently when debugging some of my tests

KatherynRizek:
I'm having issues opening and closing the chart rows for the state map tiles. I don't think it's an issue with this though since it's present on main 
Screen Shot 2022-10-31 at 2 32 53 PM
too.

flacoman91:
> I'm having issues opening and closing the chart rows for the state map tiles. I don't think it's an issue with this though since it's present on main Screen Shot 2022-10-31 at 2 32 53 PM too.

We also have this issue in Explorer see GHE issues/2140

We should go over the whole app and make sure that all clickable items have the correct cursor.  I noticed that some labels and interactive elements (especially the charts) have the pointer remain as the standard arrow.

src/components/Map/TileChartMap/TileChartMap.spec.js
flacoman91: Updated.

PR #459, opened 2022-11-17
Updated Advanced Search Tips

flacoman91:
Address GHE 1564 / CABL-230 - CCDB18

Advanced Search tips
Adding more examples to match word doc requirement
Adding mouseover tooltip
Copied tooltip css from bootstrap so we don't accidentally import other styles from bootstrap
Added footer with link to ElasticSearch

Screenshot:
Screen Shot 2022-11-18 at 9 51 55 AM

Screen Shot 2022-11-18 at 10 38 48 AM
Screen Shot 2022-11-18 at 10 38 44 AM


src/css/base.less
```diff
@@ -250,3 +250,7 @@
 @tablet-lg: ~'only screen and (min-width: 750px) and (max-width: 900px)';
 @tablet-sm: ~'only screen and (min-width: 600px) and (max-width: 749px)';
 @phone: ~'only screen and (max-width: 599px)';
+
+//tooltip
+@tooltip-background-color: @black;
+@tooltip-text-color: @white;
```
anselmbradford, line 256: Are these two used?

flacoman91: no. good catch.

src/components/Search/AdvancedTips/AdvancedTips.js
```diff
@@ -1,175 +1,288 @@
...
       <div className="o-expandable">
-        <div
-          className="o-expandable_content
-              o-expandable_content__onload-open"
-        >
-          <h3>Advanced search tips</h3>
-          <div className="tips">
-            <div className="tip">
-              <h4>And/Or</h4>
+        <div className="o-expandable_content">
```
anselmbradford, line 11: The structure here looks like it's:
```
o-expandable
  o-expandable_content
  o-expandable_content
  o-expandable_content
```

Expandables were built with the idea it would be one content area per expandable, so it is expecting this structure:

```
o-expandable
  o-expandable_content
o-expandable
  o-expandable_content
o-expandable
  o-expandable_content
```

Additionally, it would expect to have a `o-expandable_target` for toggling the expandable (see https://cfpb.github.io/design-system/components/expandables). Are these used for some expandable styles, but are not actually expandables?

flacoman91: Yeah, basically are used for the styling that comes with design system.

flacoman91: @anselmbradford 

Could you give this another look?  I'm necromancing this PR from a couple years ago.

wwhorton (approved):
Looks good to me!

PR #460, opened 2023-03-23
Add notes of hardcoded icons canonical class names

anselmbradford:
## Additions

- Add notes of hardcoded icons canonical class names

## Testing

- Only code comments, so just check for accuracy.

## Todos

- Note that some of these icons are duplicated, which could be rectified in a future PR.


src/components/iconMap.js
```diff
@@ -10,6 +11,7 @@ const iconMap = {
       <path d="M500 105.2c-276.1 0-500 223.9-500 500s223.9 500 500 500 500-223.9 500-500-223.9-500-500-500zm259 284.2L481.4 870.3c-8.2 14.1-22.7 23.4-39 24.8-1.4.1-2.9.2-4.3.2-14.8 0-28.9-6.6-38.4-18L244.4 690.9c-17.9-21-15.4-52.6 5.7-70.5 21-17.9 52.6-15.4 70.5 5.7.2.3.5.5.7.8l109.4 131.4 241.8-418.8c13.6-24 44.2-32.4 68.2-18.8 24 13.6 32.4 44.2 18.8 68.2l-.5.5z" />
     </svg>
   ),
+  // cf-icon-svg__approved-round
```
KatherynRizek, line 14: should this be `cf-icon-svg__checkmark-round`?

anselmbradford: No, thank you for asking. We are doing away with the icon "aliases" (https://github.com/cfpb/design-system/pull/1614), so any icon referred to in code should have the name from the canonical column in https://cfpb.github.io/design-system/foundation/iconography. `checkmark` is an alias of `approved` and should be referred to by its canonical name. I am leaving the "aliases" column on the iconography page only for ease of searching that page for a particular icon, but the actual duplicate icon file under the alias name will disappear. 

The reason for these comments is so that these icons show up in searches for the canonical CSS class name, however, if the hardcoded icons are updated these can go away (see below).

At some point you'll will probably want to update all these hardcoded icons with the latest icon files, which have the additional specific class name I have added in the comments in this PR, as well as changes to the viewport size (the icons used to be very large and were scaled down. Now they are closer to their intended size to begin with). Ideally you'll would have some kind of loader that injects the source SVG icon file where you need it, instead of hardcoding a duplicate of the file.

KatherynRizek: Thanks! I'll create a card on our board for when we go back to CCDB work

PR #472, opened 2023-10-04
Prevent tile map from overlapping legend on small screens

contolini:
Adds a small margin to prevent the tile map from overlapping the legend on small screens (~750px).

Fixes DATAP-30.

## Screenshots

| before | after |
|--------|-------|
| before  | after |

## Checklist

### DIFF
--- a/src/components/Map/TileChartMap/TileMap/index.js
+++ b/src/components/Map/TileChartMap/TileMap/index.js
@@ -458,11 +458,11 @@ class TileMap {
 
     // to adjust for legend height
     const mapBreakpoints = [
-      { width: 700, legendHeight: 15 },
-      { width: 580, legendHeight: 20 },
-      { width: 500, legendHeight: 30 },
-      { width: 400, legendHeight: 40 },
-      { width: 370, legendHeight: 55 },
+      { width: 700, legendHeight: 20 },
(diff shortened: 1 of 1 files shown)


src/components/Map/TileChartMap/TileMap/index.js
```diff
@@ -475,7 +475,7 @@ class TileMap {
 
     options.chart.marginRight = 0;
     options.chart.marginLeft = 0;
-    options.chart.marginTop = legendHeight;
+    options.chart.marginTop = legendHeight + 15;
```
anselmbradford, line 478: Should this be set [here](https://github.com/cfpb/ccdb5-ui/blob/main/src/components/Map/TileChartMap/TileMap/index.js#L460+L466) instead?

contolini: I assumed `legendHeight` was used to control the height of the legend but it looks like the value is only used to set a buffer between the map and the legend. So I think you're right, it probably makes more sense to alter the breakpoints array. 👍

contolini: I ended up just increasing the margin at all the breakpoints because IMHO the map is too close to the legend at most sizes.

BrewCityBoy (approved):

PR #482, opened 2023-11-09
DATAP-1208 Convert reducers to RTK slices

wwhorton:
Redux Toolkit divides redux state into "slices", which make managing state easier and more efficient. Explorer was converted to use slices, but CCDB hadn't yet. This is a PR converting all the reducers to slices, updating tests to accommodate the changes, and refactoring code where necessary, mainly in dispatches. Most of the content of the actions directory was removed since it's handled in the extraReducers piece of the slices, but where it seemed to make more sense to leave them as is, I did. 

## Removals

- boilerplate action creators
- boilerplate reducer setup
- snapshot tests
- redux-mock-store

## Changes

- numerous, but, broadly speaking, all of the reducers are now slices and have been converted in place
- the store was updated to use the newer RTK format
- added testing-library tests instead of snapshot tests
- added test setup using real reducers and actions
- added logic from Complaint Explorer
- split off logic from Query reducer into Filters, Map, and Trends reducers
- added middleware to bring together parameters to query the api
- added middleware logic to only query the aggregations endpoint only when required instead of currently querying both results and aggregations no matter what
- added React strict mode to catch more errors during development
- [Fix company typeahead](https://github.com/cfpb/ccdb5-ui/issues/531)

## Testing

- `yarn test`: All tests should pass
- Since redux touches all parts of the app, it would probably be a good idea to go through manually and make sure that everything looks as it should and works as expected.

## Notes

- The names of the tests should probably be updated to reflect the new reducer action names. It didn't occur to me until pretty late in the process, but I don't want to hold up the PR to make those updates when they could be a separate one.
- Something unexpected I learned was that when state updates in actions relied on a return value from enforceValues() they didn't actually update with the new value if it was within a returned object, e.g. { ...state, newValue: enforceValues(x,'y') }. It had to be written in the "state.value = enforceValues(x, 'y')" form. I think this might have something to do with Immer, but it's probably a good idea for all the state updates to take that form. Again, probably ought to be another PR to update that in the places it wasn't necessary for this PR.

## Todos

- Go through and prune away tests that are no longer required, and add tests to meet coverage requirements. Because the conversion alone results in an enormous PR I thought it would be a good idea to do this part as a separate PR. 
- Optimize the slices and the action directory. Again, to avoid complicating what was already a huge PR, I didn't do this, preferring to set it aside as a separate PR.

## Checklist

### DIFF
--- a/cypress/e2e/common/filters.cy.js
+++ b/cypress/e2e/common/filters.cy.js
@@ -35,7 +35,7 @@ describe('Filter Panel', () => {
       cy.intercept(request, fixture).as('getComplaintsDateTo');
 
       // eslint-disable-next-line cypress/no-unnecessary-waiting
-      cy.wait(2000);
+      cy.wait(500);
       cy.log('is expanded');
 
       cy.get('#date_received-from').should('be.visible');


src/components/Filters/Product.js
```diff
@@ -6,32 +6,32 @@ import { sortSelThenCount } from '../../utils';
...
 import { selectAggsProduct } from '../../reducers/aggs/selectors';
+import { selectFiltersProduct } from '../../reducers/filters/selectors';
+import { selectViewTab } from '../../reducers/view/selectors';
 
 /**
  * Helper function generate and sort options
  *
  * @param {Array} aggsProducts - Products array from aggs reducer
- * @param {Array} queryProduct - Products array from query reducer. TBD this will move to new filters reducer in the future
+ * @param {Array} filtersProducts - Products array from query reducer. TBD this will move to new filters reducer in the future
```
flacoman91, line 20: note to self, fix this comment.

src/components/Filters/AggregationItem/AggregationItem.spec.js
```diff
@@ -1,11 +1,11 @@
...
 import { merge } from '../../../testUtils/functionHelpers';
 import userEvent from '@testing-library/user-event';
-import * as filter from '../../../actions/filter';
+import * as filter from '../../../reducers/filters/filtersSlice';
 import * as utils from '../../../utils';
 import { slugify } from '../../../utils';
-import { defaultAggs } from '../../../reducers/aggs/aggs';
-import { defaultQuery } from '../../../reducers/query/query';
+import { aggsState } from '../../../reducers/aggs/aggsSlice';
+import { queryState } from '../../../reducers/query/querySlice';
```
flacoman91, line 8: this should be updated to filterState / filtersSlice

cdmh219:
Nothing needed from this comment, just a suggestion. Now that we're on RTK, do you think RTK Query could be a good option for our API calls? Seems like it would eliminate a lot of the custom structure we've put in place to create and make API calls. May be something to look into down the line.

flacoman91:
> Nothing needed from this comment, just a suggestion. Now that we're on RTK, do you think RTK Query could be a good option for our API calls? Seems like it would eliminate a lot of the custom structure we've put in place to create and make API calls. May be something to look into down the line.

Yes, great suggestion.  I'm all for anything that reduces custom code and not reinventing the wheel!

cdmh219:
Also doesn't look like that RTK's `createAsyncThunk()` is being used for API calls. RTK Query is a separate library and long-term solution that may or may not be a good fit, but `createAsyncThunk()` could be used immediately, as it is an out-of-the-box solution for properly integrating Redux with API calls. That will also eliminate much of our custom logic. More about this can be found [here](https://redux-toolkit.js.org/api/createAsyncThunk)

src/reducers/trends/trendsSlice.js
```diff
@@ -43,11 +49,312 @@ export const getDefaultState = () =>
...
+          },
+        };
+      },
+    },
+    dataLensChanged: {
+      reducer: (state, action) => {
+        state.subLens = '';
+        const lens = enforceValues(action.payload, 'lens');
+        switch (true) {
+          case lens === 'Company':
```
cdmh219, line 83:
```suggestion
          case 'Company':
```

src/reducers/trends/trendsSlice.js
```diff
@@ -43,11 +49,312 @@ export const getDefaultState = () =>
...
+            state.subLens = 'product';
+            break;
+          case lens === 'Overview':
```
cdmh219, line 86:
```suggestion
          case 'Overview':
```

src/reducers/trends/trendsSlice.js
```diff
@@ -43,11 +49,312 @@ export const getDefaultState = () =>
...
+            state.subLens = 'product';
+            state.chartType = 'line';
+            break;
+          case lens === 'Product':
```
cdmh219, line 90:
```suggestion
          case 'Product':
```

src/reducers/trends/trendsSlice.js
```diff
@@ -43,11 +49,312 @@ export const getDefaultState = () =>
...
+          },
+        };
+      },
+    },
+    focusChanged: {
+      reducer: (state, action) => {
+        const { focus, lens } = action.payload;
+        state.focus = focus;
+        state.lens = enforceValues(lens, 'lens');
+        state.tooltip = false;
```
cdmh219, line 162: Since this is a boolean variable, can this be changed to `hasTooltip` `isTooltip`?

src/components/ComplaintDetail/ComplaintDetail.spec.js
```diff
@@ -45,18 +45,24 @@ const renderComponent = (newDetailState, newQueryState) => {
 };
 
 describe('component::ComplaintDetail', () => {
-  it('renders loading page', () => {
+  it.only('renders loading page', () => {
```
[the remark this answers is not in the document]
flacoman91: whoops, good catch.

src/reducers/map/selectors.js
```diff
@@ -1,3 +1,4 @@
+// export const selectMapRoot = (state) => state.map;
```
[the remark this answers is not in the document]
flacoman91: Yes, I use it when I need to debug things

src/reducers/trends/trendsSlice.js
flacoman91: It's actually not a bool.  It's normally an `object` for the legend on the trends tab in the external tooltip.  

We just set this to a falsy value when resetting values

src/utils/index.js
```diff
@@ -458,3 +437,26 @@ export const selectedClass = (
...
+
+/**
+ * Remove all properties with the value 'null' from the object, or empty string
+ *
+ * @param {object} object - the object with potential nulls
+ * @returns {object} the processed object
+ */
+export function removeNullProperties(object) {
+  const keys = Object.keys(object);
+  keys.forEach((key) => {
```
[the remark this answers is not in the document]
flacoman91: I did a [quick search](https://www.google.com/search?q=you+dont+need+lodash+remove+null+properties) and found a different implementation. hopefully that's clearer.

dist/ccdb5.css
```diff
@@ -1,3 +1,3 @@
...EWtQE4zXJmClXuCWIlU5hdQxCqbqnE1KdIz79CVDvBwZxyKfQfmHTyzl01UZSvOWSTbhZLSWeDMufWLC/1ls3amT4qQq394EjIjApxBT+/nr8eEBNuKcB9SWMpmEXalNOylmlUZNTr4vE/4VdKhpC+leQf6y/e0wzL3RdJtkfUJyzwW+ZcdfgQYAJmJD3zerW6OAAAAAElFTkSuQmCC);background-size:contain;background-repeat:no-repeat}.introjs-tooltip.wide{max-width:80vw!important;width:80vw;margin-left:-40vw}.introjs-tooltip.wide .left{width:80%;float:left}.introjs-tooltip.wide .right{display:inline-block;margin-left:15px}@media screen and (min-width: 1800px){.introjs-tooltipReferenceLayer>.first-step{margin-top:-40vh!important}}@media screen and (min-width: 901px)and (max-width: 1799px){.introjs-tooltipReferenceLayer>.first-step{margin-top:-90vh!important}}@media screen and (min-width: 750px)and (max-width: 900px){.introjs-tooltipReferenceLayer>.first-step{margin-top:-110vh!important}}@media screen and (min-width: 600px)and (max-width: 749px){.introjs-tooltipReferenceLayer>.first-step{margin-top:-130vh!important}}@media screen and (max-width: 599px){.introjs-tooltipReferenceLayer>.first-step{margin-top:-110vh!important}}.tour-button{background-color:var(--pacific);height:40px;color:var(--white);border-radius:5px;position:fixed;bottom:21px;right:25px}
```
cdmh219, line 1: No changes needed to the `dist` folder, but is its content something that we really have to track or even push to 
 our repo? It can definitely be ignored via a `.gitignore` file, but I'm not sure of the actual reason we track it.

flacoman91: Yeah, it's part of the build and deploy process.  

It's how cfgov includes the assets.  All of the libraries are also in an offline cache too.

cdmh219 (approved):

PR #484, opened 2024-01-08
Fix conflict with filtered dataset export and paginated results

contolini:
We use `frm` and `search_after` query params to handle paginated search results. But when a user exports a subset of our complaint database, the aforementioned pagination params are sent to the API causing it to throw an error because [`frm mod size != 0`](https://github.com/cfpb/ccdb5-api/blob/a63efc345e1a652f9c8ec4875e9247aef6f56128/complaint_search/serializer.py#L185-L189).

Instead of reworking the API logic it's better to just purge pagination params from export URLs because they're irrelevant. The whole point of exporting is to download all the search results and not just a single page of search results.

See DATAP-33 for more info and context.

## Changes

- Updates util function that generates the export link

## Testing

- `yarn test` should pass
- `yarn cypress run --browser chrome --headless --config baseUrl=http://localhost:8000/data-research/consumer-complaints/search/` should pass
- Go to https://www.consumerfinance.gov/data-research/consumer-complaints/search/?tab=List
- Scroll to bottom and click `Next >` to get to page 2 of results.
- Scroll up and click `Export data`
- Select `Filtered dataset`
- Click `Start export`
- You should be greeted by a `400` error
- Do the above steps on localhost with this branch and it should correctly start a file download instead of 400ing

## Checklist


flacoman91 (approved):
I tested the api endpoint without those params.

I can't test it, but logically it seems solid 👍

PR #487, opened 2024-01-25
CCDB - Date range filters set to current date results in an empty chart area (#1866)

BrewCityBoy:
### Fix for Date range filters set to current date results in an empty chart area (#1866)
**Jira DATAP-1216**

This PR addresses the "0 matches" issue when the user selects the same date for the 'From' and 'Through' filter fields. 

## Changes

- adds logic for adding the 'a-text-input__error' css class for same date selection (From & Through)

## Testing

- yarn test: All tests should pass
- set From and Through filters as same date
- verify the 'a-text-input__error' css class (red border) is displayed on both input fields

## Checklist

### DIFF
--- a/src/components/Filters/DateFilter.js
+++ b/src/components/Filters/DateFilter.js
@@ -41,6 +41,7 @@ export const DateFilter = () => {
   const dispatch = useDispatch();
 
   const errorMessageText = "'From' date must be less than 'through' date";
+  const errorSameDate = "'From' date cannot be the same as 'Through' date";
 
   const fromRef = useRef();
   const throughRef = useRef();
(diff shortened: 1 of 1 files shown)


flacoman91:
Nice.  The date filter component also has some logic to display an error message when someone selects start date that appears after the end date.
https://github.com/cfpb/ccdb5-ui/blob/main/src/components/Filters/DateFilter.js#L43

Can you add to this "same date" check to that logic?   

I think leaving the message the same is fine, or to be more complete, you could explicitly say that the From and To dates cannot be the same.

BrewCityBoy:
@flacoman91 I had some logic in there, deleted it, then forgot to add it back. Nice catch and it's been added.

flacoman91 (approved):
👍  looks good to me.

PR #492, opened 2024-03-08
Remove `a-error-message`

anselmbradford:
This class is being removed in the DS in https://github.com/cfpb/design-system/pull/1926

It appears styles are attached to this class here, but it doesn't appear to be used outside of the stylesheets.

## Changes

- Remove `a-error-message` CSS class.


flacoman91 (approved):
doesn't look like this was being used anywhere.  

I would have thought that this delete icon was supposed to be red, but 🤷

anselmbradford:
> doesn't look like this was being used anywhere.
> 
> I would have thought that this delete icon was supposed to be red, but 🤷

Yeah that error message at the bottom of your screenshot is definitely missing styles, but regardless, it should be updated to  `a-form-alert` https://cfpb.github.io/design-system/components/alerts#error-1

PR #494, opened 2024-03-28
Update to design-system@0.42.0

anselmbradford:
Updates to the latest design-system, which uses CSS variables (custom properties) in place of Less variables for colors.

https://github.com/cfpb/design-system/compare/v0.39.0...v0.42.0

## Changes

- Update to design-system@0.42.0

## Testing

- The app colors should be unchanged.


src/components/Loading/Loading.less
```diff
@@ -9,17 +9,21 @@
   justify-content: center;
   align-items: center;
 
-  background-color: fade(@gray-dark, 75%);
+  // gray-dark #43484e in hex.
+  --faded-gray-dark: 67, 72, 78;
+  background-color: rgba(var(--faded-gray-dark), 0.75);
```
anselmbradford, line 14: In particular, the modal and loading background semi-opaque overlay should be unchanged.

flacoman91 (approved):
things still appear normal.

Thanks!

PR #495, opened 2024-03-29
Upgrade to design-system@0.43.0

anselmbradford:
Bring in the design-system@0.43.0 changes https://github.com/cfpb/design-system/compare/v0.42.0...v0.43.0

This change moves the `h4` font size into the expandable CSS and only makes it an `h2` size when it has the "padded" modifier.

## Changes

- Upgrade to design-system@0.43.0

## Testing

- PR checks should pass. Expandables should be unchanged.

### DIFF
--- a/src/components/Filters/__tests__/__snapshots__/DateFilter.spec.js.snap
+++ b/src/components/Filters/__tests__/__snapshots__/DateFilter.spec.js.snap
@@ -37,9 +37,9 @@ exports[`component::DateFilter snapshot shows a warning for April 2017 1`] = `
       <p
         className="u-mt15"
       >
-
+         
         CFPB updated product and issue options in April 2017 and August 2023.
-
+         
         <a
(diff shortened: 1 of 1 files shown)


src/components/Filters/CollapsibleFilter.js
```diff
@@ -60,7 +60,7 @@ export default class CollapsibleFilter extends React.Component {
           aria-label={`Hide ${this.props.title} filter`}
           onClick={this._toggleChildDisplay}
         >
-          <h3 className="h4 o-expandable_label">{this.props.title}</h3>
+          <h3 className="o-expandable_label">{this.props.title}</h3>
```
flacoman91, line 63: 👍  nice!

this h4 always looked a bit  🤔  suspect to me.

flacoman91 (approved):
👍  clean!

PR #497, opened 2024-04-05
Update hero padding

anselmbradford:
If you visit https://www.consumerfinance.gov/data-research/consumer-complaints/search/, you can see that the app has too much space above it. We should strip off the margin and padding and let the container handle that.

## Changes

- Update hero padding

## Testing

- `yarn start` and the app shouldn't have top margin and padding.


flacoman91 (approved):
looks tight 👍

PR #502, opened 2024-04-12
Update stylelint config

anselmbradford:
## Changes

- Update stylelint config to align with the config used in the design system

## Testing

- `yarn lint` should pass.


flacoman91 (approved):
Looks good!

PR #504, opened 2024-04-16
Updated to Design System 1.1.0

flacoman91:
## Additions

- BEM class linter rule for stylelint.

## Changes

- Migrates BEM syntax from old style to getBem.com style.
- Removes unneeded mega menu styles and variables.

## Testing

- PR checks should pass.


src/css/base.less
```diff
@@ -158,25 +158,25 @@
...
 
-// .grid_column__top-divider
-@grid_column__top-divider: var(--gray-40);
+// .grid__column__top-divider
+@grid__column__top-divider: var(--gray-40);
 
-// .grid_column__top-divider
-@grid_column__left-divider: var(--gray-40);
+// .grid__column__top-divider
+@grid__column__left-divider: var(--gray-40);
```
anselmbradford, line 179: Go ahead and delete any variables that aren't used, if that's any of these.

src/css/App.less
```diff
@@ -22,13 +22,13 @@
 
 // Override for Design System.
 // Fix for spacing in mega menu headings.
-.o-mega-menu_group-heading,
-.o-mega-menu_content-2-list__featured ul {
+.o-mega-menu__group-heading,
+.o-mega-menu__content-2-list--featured ul {
```
anselmbradford, line 26: Unrelated to this PR, but we shouldn't be adjusting the mega-menu from within CCDB. This should all be removed and moved to https://github.com/cfpb/consumerfinance.gov/blob/main/cfgov/unprocessed/apps/ccdb-search/css/main.less

src/components/__tests__/__snapshots__/StackedAreaChart.spec.js.snap
```diff
@@ -12,7 +12,7 @@ exports[`component: StackedAreaChart initial state renders without crashing 1`]
     className="m-notification_content"
```
anselmbradford, line 12: Do a find and replace for `m-notification_content` and update it to `m-notification__content`

I wonder if you ran the find and replace script again whether it would catch these on a second go 🤔

src/components/__tests__/__snapshots__/LineChart.spec.js.snap
```diff
@@ -12,7 +12,7 @@ exports[`component: LineChart initial state renders data lens without crashing 1
     className="m-notification_content"
```
anselmbradford, line 12: Update to `m-notification__content`

src/css/App.less
flacoman91: Can we just delete then?

src/components/Typeahead/Typeahead/Typeahead.js
```diff
@@ -32,8 +32,8 @@ export const Typeahead = ({
     <section className={`typeahead ${className | ''}`}>
       <div className="m-btn-inside-input input-contains-label">
         <div
-          className="input-contains-label_before
-                            input-contains-label_before__search"
+          className="input-contains-label__before
+                            input-contains-label__before__search"
```
anselmbradford, line 36: This is a modifier and should be updated to `input-contains-label__before--search`

package.json
```diff
@@ -74,9 +74,10 @@
     "eslint": "^8.57.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-config-react-app": "^7.0.1",
-    "eslint-plugin-cypress": "^2.15.1",
+    "eslint-plugin-cypress": "^2.15.2",
     "eslint-plugin-jsdoc": "^48.2.3",
     "eslint-plugin-react-redux": "^4.1.0",
+    "glob": "^10.3.12",
```
anselmbradford, line 80: `glob` can be removed before this goes in (after the find and replace script is removed)

src/components/__tests__/__snapshots__/StackedAreaChart.spec.js.snap
flacoman91: I tried and it did not find anything

src/css/App.less
anselmbradford: Yeah maybe so. This and below.

flacoman91: All mega menu references have been removed

src/__tests__/App.spec.js
```diff
@@ -29,7 +29,7 @@ describe('initial state', () => {
       </Provider>,
     );
 
-    expect(updateLocationHookSpy).toBeCalledTimes(1);
+    expect(updateLocationHookSpy).toBeCalled();
```
flacoman91, line 32: This hook will correctly be called twice when we update index.js to use strict mode.  This test will be rewritten in the future to remove redux-mock-store and use a real redux store.

anselmbradford (approved):

PR #506, opened 2024-05-01
restoring row expand functionality when clicking row chart names

flacoman91:
This restores row expandability when clicking a row chart name
i think an upgrade to d3 caused this functionality to break.

to test
run app
click row chart names see rows expand/collapse

There's some wonkiness when switching to Map tab, but I think it's due to the fact that chart components are still class components.

Converting them to functional components should fix this.

### DIFF
--- a/src/components/Charts/RowChart.js
+++ b/src/components/Charts/RowChart.js
@@ -196,7 +196,8 @@ export class RowChart extends React.Component {
     this.props.selectFocus(element, lens, filters);
   }
 
-  _toggleRow(rowName) {
+  _toggleRow(element) {
+    const rowName = element.target.__data__;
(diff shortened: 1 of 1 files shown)


src/components/Charts/RowChart.js
```diff
@@ -196,7 +196,8 @@ export class RowChart extends React.Component {
     this.props.selectFocus(element, lens, filters);
   }
 
-  _toggleRow(rowName) {
+  _toggleRow(element) {
+    const rowName = element.target.__data__;
```
anselmbradford, line 200: `__data__` looks like it's supposed to be private 🤷

anselmbradford (approved):

PR #508, opened 2024-05-01
DATAP-379 - convert filterPanel to functional component, bug fix Row Chart Focus

flacoman91:
- fix bug with row chart focus row element.  
- Convert FilterPanelToggle to functional component
- convert FilterPanel to functional component
- convert ExternalTooltip to functional, split off row to new component
- updated tests, remove outdated snapshot

To test:

go to trends
expand a row
click on the link "Visualize more about XYZ" 
You should get into focus mode for the Product


anselmbradford:
Not sure what it's related to this PR, but the `Visualize sub-product and issue trends for Checking or savings account >` doesn't have a hover state and it should.

anselmbradford (approved):

PR #528, opened 2024-08-15
DATAP-1494 CollapsibleFilter functional component

cdmh219:
Conversion of CollapsibleFilter from class to functional component

## Additions

-CollapsibleFilter functional component and associated unit tests

## Removals

-CollapsibleFilter class component and associated unit tests

## Changes

-References to CollapsibleFilter component in calling components

## Testing

-

## Notes

-

## Todos

-

## Checklist

### DIFF
--- a/src/components/Filters/FilterPanel.js
+++ b/src/components/Filters/FilterPanel.js
@@ -96,6 +96,7 @@ export const FilterPanel = () => {
             title="How did the consumer submit the
              complaint to the CFPB?"
             fieldName="submitted_via"
+            desc=""
           />
           <hr />
           <SimpleFilter title="Tags" desc={descTags} fieldName="tags" />


src/components/Filters/CollapsibleFilter/CollapsibleFilter.js
```diff
@@ -0,0 +1,55 @@
...
+  );
+};
+
+CollapsibleFilter.propTypes = {
+  hasChildren: PropTypes.bool,
+  className: PropTypes.string,
+  title: PropTypes.string,
+  desc: PropTypes.string,
+  children: PropTypes.node,
+};
```
flacoman91, line 53: can you add the .isRequired to these?  

```
 hasChildren: PropTypes.bool.isRequired,
  className: PropTypes.string.isRequired,
  title: PropTypes.string.isRequired,
  desc: PropTypes.string.isRequired,
  children: PropTypes.node.isRequired,
```

Can you also remove the default value from hasChildren ?  I think we should explicitly declare the value when we're using Collapsible filter. 

This will help us catch errors more easily.

In <FederalState, hasChildren is passed in as a prop, but in FilterPanel, <FederalState is used without any props.

cdmh219: This has been updated

cdmh219: Updated again to fix unit tests. Also, is className required? Based on the logic within the component that checks for its existence, it doesn't appear that it is. I've removed the `required` attribute for it, based on the previous implementation

flacoman91: You're right. className should be optional

src/components/Filters/CollapsibleFilter/CollapsibleFilter.js
```diff
@@ -0,0 +1,55 @@
...
+import { useState } from 'react';
+
+const CollapsibleFilter = ({
+  hasChildren,
+  className = '',
+  title,
+  desc,
+  children,
+}) => {
+  const [isOpen, setOpen] = useState(hasChildren);
```
flacoman91, line 13: I took another look at this, this really should just be 

const [isOpen, setOpen] = useState(true);

and we can remove hasChildren from line 7.  

Looks like all of these filters are open by default, and it really has nothing to do with whether any children exist.

cdmh219: I've updated the code to not include hasChildren. This required updates to other components and their unit tests to ensure that everything continues to work as expected

flacoman91 (approved):
Looks great! thanks!

PR #529, opened 2024-08-19
Convert trend depth toggle to functional component

flacoman91:
updated code, tests and package.json coverage thresholds.

I went ahead and did this since it already has been done in the massive RTK pull request.  I didn't want you to have to reinvent the wheel.

### DIFF
--- a/src/components/Trends/TrendDepthToggle.js
+++ b/src/components/Trends/TrendDepthToggle.js
@@ -18,7 +18,7 @@ const lensMap = {
   Company: 'company',
 };
 
-const _showMore = (filterCount, resultCount) => {
+const showMore = (filterCount, resultCount) => {
   // scenarios where we want to show more:
   // you have less visible rows that the max (5)
   if (resultCount <= maxRows) {
@@ -62,7 +62,7 @@ export const TrendDepthToggle = () => {
   }
 
   if (hasToggle) {
-    if (_showMore(queryCount, resultCount)) {
+    if (showMore(queryCount, resultCount)) {
       return (
         <div className="trend-depth-toggle">
           <button


package.json
```diff
@@ -135,10 +135,10 @@
     ],
     "coverageThreshold": {
       "global": {
-        "branches": 79,
-        "functions": 86,
-        "lines": 86,
-        "statements": 86
+        "branches": 80,
```
cdmh219, line 138: Yeah I think this can be moved back to its original values

src/components/Trends/TrendDepthToggle.js
```diff
@@ -14,63 +18,90 @@ const lensMap = {
...
+  const diff = totalResultsLength - resultCount;
+  const hasToggle = showToggle(lens, focus, totalResultsLength, queryCount);
+
+  // hide on Overview and Focus pages
+  if (focus || lens === 'Overview') {
+    return null;
+  }
+
+  if (hasToggle) {
+    if (_showMore(queryCount, resultCount)) {
```
cdmh219, line 65: Not a big deal, but would the underscore at the start of `_hasMore` be necessary here? I assumed they were present as a way to indicate they were on the instance of the TrendDepthToggle class component.

src/components/Trends/TrendsPanel.js
```diff
@@ -26,7 +26,7 @@ import Select from '../RefineBar/Select';
 import { Separator } from '../RefineBar/Separator';
 import StackedAreaChart from '../Charts/StackedAreaChart';
 import { TabbedNavigation } from '../TabbedNavigation';
-import TrendDepthToggle from './TrendDepthToggle';
+import { TrendDepthToggle } from './TrendDepthToggle';
```
cdmh219, line 29: Is there a standard around how to export these components? I've been doing `export default`, which causes it to import like the removed version of line 29, but also noticed it being exported as a variable.

src/components/Trends/TrendDepthToggle.js
flacoman91: No.  I think someone else did the underscore for functions in the past as private functions.  

Normally, I would not use them.  I'll remove them.

src/components/Trends/TrendsPanel.js
flacoman91: I prefer named exports with the brackets and I think we should stick to one convention.  Looks like there's also benefits to being explicit with the named exports.


https://react.dev/learn/importing-and-exporting-components#default-vs-named-exports
https://dev.to/phuocng/avoid-using-default-exports-a1c#:~:text=Default%20exports%20are%20used%20to,a%20function%20or%20a%20class.

We should see if we can find an eslint rule to enforce this standard.

flacoman91: Also with the named imports, you can have more than one export per file and you can unit test them more easily.

cdmh219 (approved):

PR #535, opened 2024-09-04
DATAP-1532 row chart functional component

cdmh219:
Conversion of RowChart class component to functional component

## Additions

- RowChart folder, with RowChart.js and RowChart.spec.js

## Removals

- Previous versions of RowChart.js and RowChart.spec.js

## Changes

-

## Testing

- Test coverage will be much lower for this component than others, due to D3 and Britecharts needing to be mocked out in JSDOM environment. As a result, code coverage will be much lower for component, as chart is not being rendered. 

## Notes

-

## Todos

-

## Checklist

### DIFF
--- a/src/components/Charts/RowChart/RowChart.js
+++ b/src/components/Charts/RowChart/RowChart.js
@@ -160,81 +160,73 @@ export const RowChart = ({
       }
     };
 
-    const redrawChart = () => {
-      // do this to prevent REDUX pollution
-      const rows = cloneDeep(data).filter((obj) => {
-        if (obj.name && isPrintMode) {
-          // remove spacer text if we are in print mode
-          return obj.name.indexOf('Visualize trends for') === -1;
-        }
-        return true;
-      });
-
-      if (!rows || !rows.length || !total) {
-        return;
+    // do this to prevent REDUX pollution
+    const rows = cloneDeep(data).filter((obj) => {
+      if (obj.name && isPrintMode) {
+        // remove spacer text if we are in print mode
+        return obj.name.indexOf('Visualize trends for') === -1;
       }
+      return true;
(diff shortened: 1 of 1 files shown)


src/components/Charts/RowChart/RowChart.js
```diff
@@ -0,0 +1,272 @@
...
+              .attr('y', y)
+              // eslint-disable-next-line no-mixed-operators
+              .attr('dy', ++lineNumber * lineHeight + dy + 'em')
+              .text(word);
+            wrapCount++;
+          }
+        }
+
+        // only allow this to go through if not IE
+        if (wrapCount && !window.document.documentMode) {
```
flacoman91, line 102: Wow
it's been a while since i looked at this, but we dropped IE support a long time ago.

This should be changed to 
`if (wrapCount) {`

cdmh219: Oh good catch, didn't even realize that. This has been taken out.

src/components/Charts/RowChart/RowChart.js
```diff
@@ -0,0 +1,271 @@
...
+      }
+
+      if (expandedRows.includes(rowName)) {
+        collapseARow(rowName);
+      } else {
+        expandARow(rowName);
+      }
+    };
+
+    const redrawChart = () => {
```
flacoman91, line 163: does this still work without wrapping this function with 
`const redrawChart = () => {`

I'd like to make all of these chart components look as similar as possible so it's easier to make changes

cdmh219: Yup it works without the function. Updated it as requested.

flacoman91 (approved):

PR #537, opened 2024-09-06
DATAP-1533 stacked area chart class to functional

cdmh219:
Conversion of StackedAreaChart from  class to functional  component

## Additions

-

## Removals

-

## Changes

-

## Testing

-

## Notes

-

## Todos

-

## Checklist

### DIFF
--- a/src/utils/chart.js
+++ b/src/utils/chart.js
@@ -334,7 +334,7 @@ export const pruneIncompleteStackedAreaInterval = (
   interval,
 ) => {
   const { from: dateFrom, to: dateTo } = dateRange;
-
+  //Note: this double cloneDeep is intentional to prevent state mutation in updateTrendsTooltip dispatch
   const dataClone = cloneDeep(data);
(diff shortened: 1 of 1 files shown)


src/utils/chart.js
```diff
@@ -335,17 +335,17 @@ export const pruneIncompleteStackedAreaInterval = (
 ) => {
   const { from: dateFrom, to: dateTo } = dateRange;
 
+  const dataClone = cloneDeep(data);
```
flacoman91, line 338: what kind of issues were you seeing here, and I think you may not need this

This is doing the exact same thing at L342. 

cloneDeep in our utils does the same thing too:
https://github.com/cfpb/ccdb5-ui/blob/main/src/utils/index.js#L209

flacoman91: Would this work without adding this?  This looks like it is doing the same thing twice

I would change L342 to 
let filteredData = cloneDeep(data);

flacoman91 (approved):

PR #541, opened 2024-09-13
DATAP-661 update unit tests, remove snapshot tests, accessibility fixes

flacoman91:
Addresses: DATAP-661

Added unit tests for PillPanel and TabbedNavigation
Remove remaining snapshots
Remove u-visually-hidden text from filter pills so it's more user friendly for Voiceover users:  DATAP-1541

### DIFF
--- a/src/components/Search/PillPanel.js
+++ b/src/components/Search/PillPanel.js
@@ -9,7 +9,6 @@ import {
 } from '../../reducers/query/selectors';
 
 import { useDispatch, useSelector } from 'react-redux';
-
 import dayjs from 'dayjs';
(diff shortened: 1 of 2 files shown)


src/components/Search/Pill.js
```diff
@@ -49,9 +49,6 @@ export const Pill = ({ fieldName, value }) => {
           {prefix}
           {trimmed}
         </span>
-        <span className="u-visually-hidden">
```
cdmh219, line 52: Why is this removed?

cdmh219 (approved):

PR #545, opened 2024-09-25
DATAP-1335 deprecate redux mock store

flacoman91:
Ongoing work to deprecate redux-mock-store since it's not recommended to use anymore
This will assist us when we migrate to redux toolkit

**Added**
- actions reducer to log actions called for testing
- test utility to set up store to replace redux-mock-store

**Removals**
- Remove redux-mock-store
- Remove redux-thunk

**Changes**
- moved sendHitsQuery and sendQuery into separate files in order to test and mock spies better

### DIFF
--- a/src/App.spec.js
+++ b/src/App.spec.js
@@ -25,8 +25,8 @@ describe('initial state', () => {
 
     expect(updateLocationHookSpy).toBeCalled();
 
-    expect(screen.getByText(/Consumer Complaint Database/)).toBeDefined();
-    expect(screen.getByText(/Search within/)).toBeDefined();
+    expect(screen.getByText(/Consumer Complaint Database/)).toBeInTheDocument();
+    expect(screen.getByText(/Search within/)).toBeInTheDocument();
     expect(
       screen.getByRole('button', { name: /Show advanced search tips/ }),
     ).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


src/App.spec.js
```diff
@@ -1,26 +1,41 @@
...
+
+    render(
+      <Provider store={store}>
+        <App />
+      </Provider>,
+    );
+
+    expect(updateLocationHookSpy).toBeCalled();
+
+    expect(screen.getByText(/Consumer Complaint Database/)).toBeDefined();
```
cdmh219, line 28: Not a huge deal, but are we able to use the `toBeIntheDocument()` assertion here instead of `toBeDefined()`? It's a bit more descriptive of what's actually being tested. This would also apply in other tests where `toBeDefined()` is used.

src/middleware/actionLogger/actionLogger.spec.js
```diff
@@ -0,0 +1,13 @@
+import { setupStore } from '../../testUtils/setupStore';
+
+describe('redux middleware::actionLogger', () => {
+  test('it logs actions', () => {
```
cdmh219, line 4: Should this be `test()` instead of `it()` (like all other tests)?

src/App.spec.js
flacoman91: You're right. that is more correct.

src/middleware/actionLogger/actionLogger.spec.js
flacoman91: I believe it's an alias and they're interchangeable, but its better to be consistent.

cdmh219 (approved):

PR #549, opened 2024-09-27
DATAP-1550 tileChartMap prevent state mutation

flacoman91:
Update dependencies
Adding cloneDeep function to tileChartMap so it doesn't pollute redux state
to test:

yarn install
yarn start
toggle Map tab
confirm there is no warning that appears like so:

Screenshot 2024-09-27 at 2 22 51 PM


cdmh219 (commented):
Looks good, but also looks like there may be failing tests. Please take a look and update accordingly.

flacoman91:
> Looks good, but also looks like there may be failing tests. Please take a look and update accordingly.

I kicked off the FE test in CI and they're passing now.  We should probably rewrite those tests to be more robust.

cdmh219 (approved):

PR #552, opened 2024-10-07
DATAP-1208 - update aggregations reducer to aggsSlice

flacoman91:
Part 1 of many to convert redux reducers to redux slices


flacoman91:
Going to close this and just focus on PR #482  

There's quite a bit that would need to be backported and it's extra work that isn't necessary since it would be removed anyway.

PR #553, opened 2024-10-24
Migrate to RTK Query

flacoman91:
This updates the app to use RTK Query


src/reducers/query/querySlice.spec.js
```diff
@@ -43,85 +41,84 @@ describe('reducer:query', () => {
...
-          4: [4, 2],
-          5: [5, 2],
-        },
-        page: 100,
-        size: 100,
-        totalPages: 5,
-      });
-    });
-  });
+  // describe('COMPLAINTS_RECEIVED actions', () => {
```
cdmh219, line 49: Can this commented code be deleted?

src/components/Trends/TrendDepthToggle/TrendDepthToggle.spec.js
```diff
@@ -0,0 +1,192 @@
...
+    });
+    renderComponent(
+      {},
+      {
+        lens: 'Product',
+        trendDepth: 14,
+      },
+    );
+
+    await screen.findByRole('button', { name: 'Show less' });
```
cdmh219, line 118: Can button be used in one variable across both lines 118 and 119?

src/components/Trends/TrendDepthToggle/TrendDepthToggle.spec.js
```diff
@@ -0,0 +1,192 @@
...
+        });
+      } else if (req.url.indexOf('API?') > -1) {
+        // this is the list
+        return Promise.resolve({
+          body: JSON.stringify(aggResponse),
+        });
+      }
+    });
+    renderComponent({}, { focus: 'Foo bar' });
+    expect(screen.queryByRole('button', { name: 'Show more' })).toBeNull();
```
cdmh219, line 64:
```suggestion
    expect(screen.queryByRole('button', { name: 'Show more' })).not.toBeInTheDocument();
```

src/components/Trends/TrendDepthToggle/TrendDepthToggle.spec.js
```diff
@@ -0,0 +1,192 @@
...
+  });
+
+  it('does not render when lens is Overview', () => {
+    renderComponent({}, { lens: 'Overview' });
+    expect(screen.queryByRole('button', { name: 'Show more' })).toBeNull();
```
cdmh219, line 69: Could be updated like line 64

src/api/hooks/useGetAggregations.js
```diff
@@ -0,0 +1,16 @@
+import { useGetAggregationsQuery } from '../complaints/complaints';
```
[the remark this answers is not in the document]
flacoman91: Yeah, i thought about this, I originally had it under the hooks folder, but I moved them here to keep them closer to the RTKQ API

src/api/hooks/useGetAggregations.js
```diff
@@ -0,0 +1,16 @@
...
+import { useSelector } from 'react-redux';
```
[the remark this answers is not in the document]
flacoman91: Yes we can. 

They're already implicitly being called through the rendering of the components.

I found that out the hard way  🤦 , 

you'll see a lot of duplicate dummy route reducer and tab view values in the components to let the hooks fire.

src/api/complaints/complaints.js
```diff
@@ -0,0 +1,114 @@
+import { createApi, fetchBaseQuery } from '@reduxjs/toolkit/query/react';
```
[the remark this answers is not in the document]
flacoman91: Yes,we don't need the folder.  

I was debating on if we should split the api up (like our reducers) and then use [injectEndpoints](https://redux-toolkit.js.org/rtk-query/usage/code-splitting) to put them back together, but our api here isn't that complicated to need that.

src/components/Trends/TrendDepthToggle/TrendDepthToggle.spec.js
flacoman91: This is intentional, this case, await returns a Promise and we use it here to allow the button to render asynchronously. 

Normally in Cypress tests we're not supposed to be saving the selectors in a variable.  They don't recommend it.

> This is one of the primary reasons why we do not recommend using [variables](https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Variables) within your tests.

https://learn.cypress.io/cypress-fundamentals/understanding-the-asynchronous-nature-of-cypress

Typically I prefer to have the screen.getBy calls repeated explicitly instead of saving it in a variable  because if the DOM somehow changes, each screen.getBy call is a new call to check the document instead of having a saved variable.

src/reducers/query/querySlice.spec.js
flacoman91: That reminds me, I commented out these tests since these need to be moved over to validate the API.

src/utils/chart.js
```diff
@@ -228,13 +229,27 @@ export const updateDateBuckets = (name, buckets, areaBuckets) => {
...
 };
 
-export const externalTooltipFormatter = (tooltip) => {
+export const externalTooltipFormatter = (tooltip, colorMap) => {
   if (!tooltip) {
     return tooltip;
   }
+  const newTooltip = { ...tooltip };
   const parts = tooltip.title.split(':');
+  newTooltip.values = cloneDeep(tooltip.values);
```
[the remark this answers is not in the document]
flacoman91: That was giving an error since tooltip.values is an Array.  I went ahead and switched it over to structuredClone and added a global polyfill in setupTests.js

flacoman91:
> In addition to the feedback provided in each file, please also make sure to update the Cypress tests accordingly (if needed). When I tried running `yarn run cypress run` on this branch, I got errors that implied the updated REST calls weren't being recognized. Please verify on your end.

The tests were updated since some API calls changed. 

I wasn't seeing this.  I can look into it further if you can provide a screenshot.  

The cypress tests are also running in CI.

src/utils/index.js
```diff
@@ -456,4 +457,23 @@ export function removeNullProperties(object) {
     }
     return acc;
   }, {});
+
+  for (const key in myObject) {
```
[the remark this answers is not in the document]
flacoman91: Thanks for the suggestion.  I took a look at this, and its a bit harder for me to parse through. 
i made this fiddle here https://jsfiddle.net/72rgut53/ to see if it would work, and we'd still need to delete the key from the object.

src/components/Filters/Aggregation/AggregationItem/AggregationItem.spec.js
```diff
@@ -83,69 +75,70 @@ describe('component::AggregationItem', () => {
...
       const ownProps = {
         fieldName: 'issue',
         item: { key: 'No Money', doc_count: 1 },
       };
-
-      renderComponent(ownProps, aggs, filters);
-
+      renderComponent(ownProps, filters);
+      await screen.findByRole('checkbox');
       expect(screen.getByRole('checkbox')).not.toBeChecked();
```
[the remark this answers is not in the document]
flacoman91: We're not supposed to do that, but technically we can use aliases for querying and retryability

https://docs.cypress.io/app/core-concepts/introduction-to-cypress#Using-Aliases-to-Refer-to-Previous-Subjects

src/api/params/params.spec.js
```diff
@@ -49,64 +56,90 @@ describe('api.v2.params', () => {
...
-    describe('parseParamsToQuery', () => {
-      it('handles bogus searchFieldMap', () => {
-        fixtureStore.query.searchField = 'Bogus value';
-        const actual = sut.extractQueryParams(fixtureStore.query);
-        expect(actual.field).toEqual('all');
-      });
+    it('handles bogus searchFieldMap', () => {
+      fixtureStore.query.searchField = 'Bogus value';
+      const actual = sut.extractQueryParams(fixtureStore.query);
+      expect(actual.field).toEqual('all');
```
cdmh219, line 128: I would steer clear of using `toEqual()` clauses for evaluating primitive values, as it's more for object/array values. `toBe()` would be the way to go.
https://jestjs.io/docs/expect#tobevalue
https://jestjs.io/docs/expect#toequalvalue

```suggestion
      expect(actual.field).toBe('all');
```

src/api/params/params.spec.js
```diff
@@ -49,64 +56,90 @@ describe('api.v2.params', () => {
...
-        fixtureStore.query.size = 10;
-        const actual = sut.extractQueryParams(fixtureStore.query);
-        expect(actual.frm).toEqual(0);
-      });
+    it('calculates frm from page and size when from is not given', () => {
+      delete fixtureStore.query.from;
+      fixtureStore.query.page = 0;
+      fixtureStore.query.size = 10;
+      const actual = sut.extractQueryParams(fixtureStore.query);
+      expect(actual.frm).toEqual(0);
```
cdmh219, line 136:
```suggestion
      expect(actual.frm).toBe(0);
```

src/api/params/params.spec.js
```diff
@@ -49,64 +56,90 @@ describe('api.v2.params', () => {
...
 
-      it('assigns query.searchAfter to search_after', () => {
-        fixtureStore.query.searchAfter = '1__12345';
-        const actual = sut.extractQueryParams(fixtureStore.query);
-        expect(actual.search_after).toEqual('1__12345');
-      });
+    it('assigns query.searchAfter to search_after', () => {
+      fixtureStore.query.searchAfter = '1__12345';
+      const actual = sut.extractQueryParams(fixtureStore.query);
+      expect(actual.search_after).toEqual('1__12345');
```
cdmh219, line 142:
```suggestion
      expect(actual.search_after).toBe('1__12345');
```

cdmh219 (approved):

cdmh219 (changes requested):
Please see previous comments on the use of  toBe() vs toEqual() in tests. Otherwise, everything else looks good

src/api/params/params.spec.js
flacoman91: thanks! i didnt know that.  I went ahead and updated our tests where we test primitives to use toBe

PR #578, opened 2025-04-14
DATAP-1691- date filter fix

flacoman91:
- This fixes a bug in the date input that occurs only in Chrome and Edge where when you type in the day starting with 0, for instance 01-04-2021 the browser resets the date to the minimum date.
- Updated package dependencies
- Updated Cypress test

**To test:**
run code locally or go to dev2

Test the app with the following browsers and confirm the date filters work correctly:
Chrome, Edge, Safari

### FROM date validation

**Validate the From date filter works and that typing a day starting with the number 0 does not cause the date filter to reset**
Open Date filter
In the From date:
Type:  04 {tab} 05 {tab} 2020 {tab}
Confirm the from Date is 04/05/2020 and applied to the site

**Validate that entering an incorrect date will reset the date to the Min date 12/01/2011"**
Open Date filter
In the From date:
Type:  04 {tab} 05 {tab} 0001 {tab}
Confirm the from Date is 12/01/2011 and applied to the site

### THROUGH date validation

**Validate the Through date filter works and that typing a day starting with the number 0 does not cause the date filter to reset**
Open Date filter
In the From date:
Type:  04 {tab} 05 {tab} 2024 {tab}
Confirm the through Date is 04/05/2024 and applied to the site

**Validate that entering an incorrect date will reset the date to the Max date 4/14/2025 (or today's date)**
Open Date filter
In the Through date:
Type:  04 {tab} 05 {tab} 0001 {tab}
Confirm the through Date is 04/14/2025 and applied to the site


higs4281 (approved):
I leave code critique to my betters, but I follow it, and functionality in DEV2 is good..

cypress/e2e/common/filters.cy.js
```diff
@@ -84,6 +88,15 @@ describe('Filter Panel', () => {
       cy.url().should('include', 'date_received_max=2020-10-31');
 
       cy.wait('@getComplaintsDateTo');
+
+      // check error handling and default values
+      cy.get('#date_received-from').type('2000-09-11', {
+        force: true,
```
anselmbradford, line 94: Is the `force: true` needed? Would `delay: 0` be an option?

flacoman91:
> The test steps worked as described, however, there's one quirk I found:
> 
> * Enter the FROM incorrectly and tab out of the field. The app performs a search and resets to 12/01/2011.
> * Enter the THROUGH incorrectly with 04/05/0001. The app shows the `'From' date must be less than 'through' date` validation error. Tab outside of the THROUGH field and the app searches and resets to today's date. The validation error is cleared.
> * Enter the THROUGH incorrectly with 04/05/0001 a second time, and tab outside the field. The field resets to today's date, but the validation error gets stuck in a shown state, as the app doesn't need to perform a new search to clear it.
> 
> Do you want to address that here, or in a new issue?

thanks for the thorough review! I'll take care of it here and let you know when its ready to retest.
thanks!

cypress/e2e/common/filters.cy.js
flacoman91: let me check this and get back to you

flacoman91: Turns out it wasn't needed.  I cleaned up a bunch of the force true options

PR #582, opened 2025-06-10
DATAP-1711 - fix Through Date mismatch

flacoman91:
The through date was using a value from [metadata.js ](https://files.consumerfinance.gov/ccdb/metadata.js)

That value is inaccurate and appears to be 1 day behind the value coming from the API.

- I've updated the app to wait for an initial call to get the value from the API and set the MAX_DATE value.
- Fix a react warning indicating that key should not be set through spread operators
- Update unit tests
- Simplified Cypress tests to not watch for network request and wait for Loading screen to end instead

To test:
Open network request tab in chrome devtools
Go to dev5
confirm you don't see any calls to metadata.js

Everything should work the same as usual


flacoman91:
This is on dev5 for review

PR #583, opened 2025-06-12
DATAP-1750 DATAP-1752 es sanitization, remove summary element

flacoman91:
Addresses: DATAP-1752, DATAP-1750
This PR:
- removes Summary element to fix AXE warning
- removes ES sanitization so that users can type in whatever they want in the text box

Restructure some components that were causing mount/remount on api call fail which turned into an infinite loop. 

To verify:
Open Network tab in Chrome
Search for an invalid query such as: `1/4`
You should only see 2 requests occurring and failing. (on production, it goes in an infinite loop)

Valid queries should still work
`product:("Debt collection")`

This is on dev5 for testing


higs4281 (approved):
results confirmed locally

PR #590, opened 2025-08-14
DATAP-897 - create react app to rsbuild migration

flacoman91:
Addresses: DATAP-897

- Updated build script to copy font file from node_modules
- Consolidate flaky cypress filter test due to it failing constantly in CI (but not locally) 🤔 
- Consolidate other cypress tests to take advantage of RTK Query caching to make testing faster
- Removes create react app in favor of rsbuild
- browserslist-rs currently doesn't support queries, so we are just using the standard rules. [link](https://github.com/browserslist/browserslist-rs/issues/37). Once this is supported we will switch to use cfpb shared configs.

To test:
delete node_modules

yarn commands should still work

yarn install;
yarn test;
yarn build;
yarn start

### DIFF
--- a/cypress/e2e/common/search-bar.cy.js
+++ b/cypress/e2e/common/search-bar.cy.js
@@ -9,32 +9,26 @@ describe('Search Bar', () => {
     '**/data-research/consumer-complaints/search/' +
     'api/v1/_suggest_company/**';
 
-  beforeEach(() => {
-    cy.visit('?tab=List');
-    waitForLoading();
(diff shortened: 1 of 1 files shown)


flacoman91:
> After getting node 24 running locally, I could run all the suggested yarn commands except for `yarn serve`, which did nothing. However `yarn start` did work, and fired up a ccdb instance at port 3000.

Whoops, you are correct, I meant `yarn start`

PR #598, opened 2026-02-03
Upgrade yarn, update package deps, fix dependabot errors

flacoman91:
This updates the repo so it's not so outdated.

- Update to use yarn modern. 
- update CI commands
- Update NPM packages
- swap CFPB DS for CFPB DSR
- fix linting warnings
- Update DS references in scss files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,11 +14,9 @@ jobs:
         with:
           node-version: '24.x'
 
-      - name: Enable Corepack
-        run: corepack enable
-
       - name: Install Node dependencies
         run: |
+          corepack enable
           yarn install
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -14,10 +14,16 @@ jobs:
         with:
           node-version: '24.x'
 
+      - name: Enable Corepack
+        run: corepack enable
+
```
anselmbradford: This is fine, but you could also merge this in with the install step if you want it slightly more succinct (see https://github.com/cfpb/design-system/blob/fcc3df5c0cd9b59a975fc67ec81bdcb7959a3b01/.github/actions/shared-action-setup/action.yml#L18)

PR #619, opened 2026-07-28
DATAP-2042 - Remove JSON format

flacoman91:
## Summary

Updates the complaint data download experience and related UI/tooling:

- Removes JSON as a download option; filtered and full downloads are CSV (full dataset remains a zipped CSV).
- Redesigned “Download complaint data” modal, clearer dataset options, and filter/limit error alerts when filtered results can’t be downloaded.
- Prefers filtered results when filters are applied and under the 100k limit; otherwise defaults to all data and disables filtered when it’s empty or over the limit.
- Copy-link polish - Saves the current results URL; success state uses the secondary button (“Link copied” + check icon) instead of the old green style; input/button join without a gray shared edge.
- Page-loading overlay - Blocks interaction (including Export) while aggregations/results are refetching so the modal doesn’t open with empty counts.
- Filter panel layout - Removes extra side padding on expandables and the narratives checkbox so controls use the full panel width.
- DS/DSR cleanup -  Drops redundant local styles covered by the design system; bumps DSR; small tab spacing tweak.
- TypeScript kickoff - Adds Babel TS + `yarn typecheck`; converts leaf modules (`breakpoints`, `colors`, `compare`, `format-date`); prunes unused utils/constants.

## Test plan

- [ ] Open Export with no filters → filtered option disabled + empty-filter alert; All selected
- [ ] Apply filters under 100k → Filtered selected and enabled; download CSV
- [ ] Apply filters over 100k → Filtered disabled + limit alert; All selected
- [ ] Copy link → “Link copied” secondary button; URL is the page URL
- [ ] Confirm modal title/copy and Download/Cancel/Close actions
- [ ] While results are loading, Export is blocked by the page overlay
- [ ] Filter panel expandables and narratives checkbox align full-width
- [ ] Unit tests / typecheck pass

PR #626, opened 2026-07-30
List view clean up

flacoman91:
- Cleanup List view
- Removed custom selects in favor of standard DSR selects
- Removed disabled options in favor of filtering out invalid options
- Updated Filter panel options
- Removed export modal since it doesn't seem to be required since there are no options.

To verify, validate against figma mockups:


flacoman91:
putting this on hold since we're rolling out #619  first

natalia-fitzgerald:
@flacoman91 
Can you update this PR to include the latest updates to the filter panel, the complaint card, and the search well (date format)?

PR #630, opened 2026-08-14
Remove complaint narratives, default to List view

wpears:
Per stakeholder direction, this PR removes consumer complaint narratives from the frontend along with the map and trends tabs/views.

## Summary

Results are List-only now. Map, Trends, and complaint narratives are gone from the UI and from leftover code.

- Removed tabbed navigation from the results panel and made List the only view.
- Removed narratives: `has-narrative` filter, narrative buttons, and narrative text on cards/detail.
- Deleted leftover Map/Trends/charts code (panels, reducers, API hooks, tour steps, Cypress).
- Dropped unused charting deps (`highcharts`, `d3`, `britecharts`).
- Bookmarks with `tab=List` still work. Map/Trends URL values coerce to List.

Also started a small TypeScript conversion on the Redux read path (selectors, store, a few hooks) so we can migrate incrementally.

## Test plan

- [ ] Search results load in List view with no Map/Trends tabs
- [ ] Filters, pagination, export, and complaint detail still work
- [ ] Old `?tab` URLs fall back to List
- [ ] Tour still runs and only describes List
- [ ] No narrative filter or narrative text on cards/detail


flacoman91 (commented):
Can you remove the britecharts, and high charts deps from package.json? I think there may be some cypress tests, babel or rsbuild configs you can remove too

wpears:
>Can you remove the britecharts, and high charts deps from package.json? I think there may be some babel or rsbuild configs you can remove too

Yep, this first pass was more or less me trying to introduce the change as gently as possible before you looked at it. Let's chat Monday when everyone's back.

flacoman91 (approved):
LGTM... got a bunch of TS updates incoming after this one

PR #633, opened 2026-08-28
Remove unnecessary useMemo and useCallback wrappers

flacoman91:
Turns out we were abusing useMemo and using it incorrectly!  This cleanup matches the cleanup in Explorer

## Summary

React `useMemo` / `useCallback` only pay off when they keep a stable identity that a memoized child or an effect actually depends on, or when they skip genuinely expensive work. Most of the hooks we had were wrapping cheap booleans, string builds, and JSX, or wrapping callbacks that `TourSteps` never compared by identity. That added dependency arrays and extra memo bookkeeping without preventing any child work.

This PR drops those wrappers and keeps the few cases that still matter:

- **Tour `baseSteps` `useMemo`** — `TourSteps` reconfigures intro.js when `prevProps.steps !== steps`. A new array every render would restart the tour.
- **Typeahead `onSearchHandler` `useCallback`** — zip/company filters do not pass `handleChange`, so a stable handler is what keeps react-bootstrap-typeahead’s 250ms `onSearch` debounce from resetting. Cargo-cult `setState` deps were removed.
- **Imperative `useRef`s** — typeahead `.clear()`, date-input `.blur()`, tour `introJs` instance. Those are not memoization.

What changed:

- List panel: `width < 750` is a boolean, not a memo.
- Search bar: pass the module-level `searchFieldOptions` constant instead of memoizing it.
- Print info, complaint detail, data export, refine-bar select: cheap strings/JSX/option maps computed inline.
- Root modal: hoisted a `MODALS` lookup map.
- Date filters: error text and input class names moved to small helpers (also keeps the complexity cap).
- Tour: dropped four `useCallback`s and the extra `handleX` → `onX` wrappers. Intro.js reads latest props when events fire.

## Test plan

- [ ] `yarn test` — list panel, search bar, print info, complaint detail, root modal, data export, select, async typeahead, date filters, tour
- [ ] Search well: field select, text search, company typeahead suggestions + select
- [ ] Date received and company-received filters: invalid range still shows errors and error input classes
- [ ] List refine bar: size and sort dropdowns
- [ ] Complaint detail back link
- [ ] Export modal: filtered vs all download URLs
- [ ] Tour: desktop steps; mobile filter open/close steps still advance
- [ ] Print view complaint counts

### DIFF
--- a/cypress/e2e/common/filters.cy.js
+++ b/cypress/e2e/common/filters.cy.js
@@ -39,8 +39,7 @@ const sizeSelect = () => cy.findByLabelText('Show per page');
 
 const complaintLinks = () => cy.findAllByRole('link', { name: /^Complaint / });
 
-const filterPills = () =>
-  cy.findByRole('heading', { name: 'Filters applied:' }).closest('section');
+const filterPills = () => cy.findByText('Filters applied:').closest('section');
 
 describe('Filter Panel', () => {
   it('allows the app to filter complaints', () => {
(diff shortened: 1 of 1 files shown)


wpears (approved):
Nice simplifications! Still some cypress stuff to clean before merge though :)

flacoman91:
Going to merge this in.   @natalia-fitzgerald if you have any ui updates you can review what is most current at https://cfpb.github.io/ccdb5-ui/

PR #634, opened 2026-08-28
Add react compiler

wpears:
After #633, I think it's worth adding [react compiler](https://react.dev/learn/react-compiler/introduction). We can then probably strip out more manual useMemo calls.


flacoman91:
Do we need to enable this if we're already using the included [rsbuild plugin](https://github.com/cfpb/ccdb5-ui/blob/main/rsbuild.config.mts#L92)?  https://rsbuild.rs/guide/framework/react#react-compiler

wpears:
>Do we need to enable this if we're already using the included [rsbuild plugin](https://github.com/cfpb/ccdb5-ui/blob/main/rsbuild.config.mts#L92)? https://rsbuild.rs/guide/framework/react#react-compiler

Oh! No we don't! Interesting, so we can already kill a bunch of manual memoing

PR #636, opened 2026-08-31
Replace jest with rstest and remove babel

wpears:
Babel was mostly vestigial; removing jest makes it entirely so.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,11 @@ jobs:
       - name: Run front-end unit tests
         run: yarn run test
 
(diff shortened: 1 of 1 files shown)


flacoman91:
Since we are still using coveralls, should we restore the removal 

https://github.com/cfpb/ccdb5-ui/pull/636/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL45-L47

flacoman91 (approved):
Nice!  Going to have to some of this and get it into explorer!

PR #637, opened 2026-09-03
replace babel jest

flacoman91:
updates rstest coverage so coveralls continues to work


flacoman91:
@wpears can you take a look at this and see if it makes sense to keep coveralls going?

PR #639, opened 2026-09-09
Add export route to allow linking to the full download button

wpears:
This adds a simple little component (attached screenshot) based on the data export dialog.
This component is routable to at `/export/` so we can link to it from, eg, the wagtail landing page.

I believe *most* of the crawling/scraping of the full export file is due to it being hard linked on the ccdb landing page. Instead linking to this export route with a button but no links will hopefully foil the more naive scrapers.

Screenshot 2026-09-09 at 1 32 24 PM


flacoman91 (approved):
looks good to me.  It doesn't have any impact on the current site and provides a placeholder that wagtail can link to.

PR #640, opened 2026-09-10
Dsr alert

flacoman91:
### Changes include
- Updates CCDB to use the CFPB Design System alert component (for the results count).
  - This is approach is consistent with Complaint Explorer. 
- Changes "Export data" link text to "Download data" 
  - This is consistent with the rest of the user flow (modal language, supporting pages, etc.)

**Screenshots**
| Desktop | Mobile |  
|------------|------------|
|Screenshot 2026-09-10 at 10 55 46 AM|Screenshot 2026-09-10 at 10 46 12 AM|


natalia-fitzgerald:
@flacoman91 
In the absence of an open ticket, I am logging CCDB bugs I notice here for the time being. 
- [ ] The icon is sitting high in the alert. I think the issue may be coming from the DS? The icon should sit vertically centered. I may need to log this on the DS side but putting it here so I don't lose it

PR #646, opened 2026-09-14
update ui

flacoman91:
- [x] At desktop, in the results count alert the vertical padding between the count and the links should be 8px (at desktop), In staging it is set to 15px. 
- [x] At mobile, the vertical padding between filter items should be 45px 
- [x] At mobile, the space at the bottom of the page before the footer should be 60px. 
- [x] At desktop, the vertical padding between filter items in the filter panel is 45px (this is correct on staging)

Screenshot 2026-09-14 at 12 52 16 PM
Screenshot 2026-09-14 at 12 52 22 PM


natalia-fitzgerald (approved):
@flacoman91 Nice work!

